PR #2, opened 2021-04-02
Initial Pixel Area support.

PaulHuwe:
Added files to support pixel area, and took a stab at integrating that into a dependent object (wfi_image). 
Please give me feedback on this draft. I have no tests nor documentation for this in RAD - I am not sure where to do that at present. I know I will need to update CHANGES.rst, but I am waiting until I get feedback, as the other WFI object modules will need updating as well for a real PR.


src/rad/resources/schemas/pixelarea-1.0.0.yaml
```diff
@@ -0,0 +1,27 @@
...
+      datatype:
+        enum: [float32]
+      shape:
+        type: array
+        minItems: 2
+        maxItems: 2
+    required: [datatype, shape]
+propertyOrder: [meta, area]
+flowStyle: block
+required: [meta, area]
```
perrygreenfield, line 27: Perhaps I'm wrong, but I don't think meta needs to be part of this since it is already defined in the top level schema (wfi_image).

src/rad/resources/schemas/wfi_image-1.0.0.yaml
```diff
@@ -76,17 +76,18 @@ properties:
...
-    tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
-    type object:
-    properties:
-      datatype:
-        enum: [float32]
-      shape:
-        type: array
-        minItems: 2
-        maxItems: 2
+    required: [datatype, shape]
```
perrygreenfield, line 79: Since the value here is an object (what type we can think about and discuss) it should be using a tag here I think. Something like 
`
area: "tag:stsci.edu:datamodels/roman/pixelarea-1.0.0"

The trick here is that one desires a numpy array as the value, but one also wants the machinery to recognize that this is a pixelarea object so that when writing out, it should be so tagged. I think we can return a subclass of a numpy array that changes nothing about it other than gives it a distinct class so that it will be properly tagged. I'll ask Ed to see if he sees any problems with that. If it changes no behavior I don't think it should cause any problems but I would want to see that when operating with other numpy arrays (e.g.,, multiplying them) that they come out as ordinary numpy arrays. So I'd like to do a little testing of what happens there. It should be simple enough to test.
`

perrygreenfield: Based on my initial tests it is starting to look like the $ref approach is the right one. Instead of doing it the way you had the key name part of the $ref. Ed tells me that it is possible to do this:

```
%YAML 1.1
---
$schema: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
id: "http://stsci.edu/schemas/roman/wfi_image-1.0.0"
title: |
  The schema for WFI level 2 images
type: object
properties:
  meta:
    allOf:
      - $ref: "http://stsci.edu/schemas/datamodels/roman/common-1.0.0"
<stuff removed>
  area:
    $ref: "http......pixelarea-1.0.0"
```

perrygreenfield: Ed pointed out there are some special customizations for numpy arrays that greatly simplify the schemas. So the resulting schema to be used would be:

```
%YAML 1.1
---
$schema: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
id: "http://stsci.edu/schemas/roman/pixelarea-1.0.0"

title: |
datatype: float32
ndim: 2
```

src/rad/resources/schemas/pixelarea-1.0.0.yaml
PaulHuwe: Meta removed

src/rad/resources/schemas/wfi_image-1.0.0.yaml
PaulHuwe: Implemented.

src/rad/resources/schemas/keyword_pixelarea-1.0.0.yaml
```diff
@@ -0,0 +1,23 @@
+%YAML 1.1
```
perrygreenfield, line 1: Is the designation keyword appropriate here since it has multiple attributes? It looks more like an object to me. I had interpreted the JWST keyword schemas as pertaining to simple keyword/value pairs. Though I do see fro this one it does the same thing. But I wonder if it shouldn't simply be added to the photometry schema, both of these values are already present, so this doesn't look like it is necessary, but I may be missing something.

src/rad/resources/schemas/pixelarea-1.0.0.yaml
```diff
@@ -0,0 +1,12 @@
+%YAML 1.1
```
perrygreenfield, line 1: I wonder if we shouldn't just call this a float image schema since nothing prevents it from being reused for all floating point arrays of 2 dimensions. And a DQ one for data quality "images",

src/rad/resources/schemas/keyword_pixelarea-1.0.0.yaml
PaulHuwe: I agree - I think this file is extraneous and can be removed. I see no reference to it in jwst code outside of schema files.

src/rad/resources/schemas/pixelarea-1.0.0.yaml
PaulHuwe: Hrmm, actually I am wondering if we should go back to my initial ref command idea, instead of having a tag under area [in wfi_image]. (i.e.
This:
  $ref: "http://stsci.edu/schemas/datamodels/roman/pixelarea-1.0.0"
Instead of:
  area:
    $ref: "http://stsci.edu/schemas/datamodels/roman/pixelarea-1.0.0"
) And I would adjust pixelarea accordingly.

The reason to go back to what I had is that in going down the route we have been, we are eliminating the PixelAreaModel Datamodel object as a distinct item. I don't see it used extensively in the code, but in changing it, we also lose the objects sharing.

What do y'all think?

eslavich: I'm not sure I understand the idea, are you suggesting we define a class like this?

```python
class PixelAreaArray(np.ndarray):
  pass
```

or maybe like this?

```python
class PixelAreaModel:
  def __init__(self, array):
    self.array = array
 ```

PaulHuwe: Oh, as in undoing the class specification and going back to defining the area variable within pixelarea and then importing it into wfi_image.

eslavich: Ah ha I didn't realize that there was a pixel area reference type, I thought we were just adding a reusable component for a 2D float array.  @perrygreenfield doesn't that mean we need a dedicated tag?

perrygreenfield: I'm not sure. Pixel area should be directly derivable from the WCS if I'm understanding it correctly.

PaulHuwe: Right, well, I think we need to decide to either:
1. Preserve the datamodel + reference type.
2. Just leave area as a variable in wfi_* alone (like data, etc.) and delete pixelarea schema.

What do y'all think? Other voices want to chip in? (@dmggh @ddavis-stsci @nden @sosey ?)

src/rad/resources/schemas/keyword_pixelarea-1.0.0.yaml
PaulHuwe: Removed.

src/rad/resources/schemas/pixelarea-1.0.0.yaml
PaulHuwe: I decided that this should remain as a datamodel + reference. I don't want to change much in rad that requires lots of downstream changes in other packages, and removes the plug & play aspect of what we are building.

PR #15, opened 2021-04-20
added  test

perrygreenfield:
Added a test to check for schema validity. Also updated .gitignore to filter out .DS_Store files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ MANIFEST
 #  before PyInstaller builds the exe, so as to inject date/other infos into it.
 *.manifest
 *.spec
+.DS_Store
 
 # Installer logs
 pip-log.txt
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -31,6 +31,7 @@ MANIFEST
 #  before PyInstaller builds the exe, so as to inject date/other infos into it.
 *.manifest
 *.spec
+.DS_Store
```
eslavich, line 34: OS-specific files are typically left out of the project's .gitignore and added to the user's .gitignore_global instead

src/rad/tests/test_extension.py
```diff
@@ -0,0 +1,19 @@
+import yaml
+import asdf
+
+from rad.integration import get_resource_mappings
+
+
+def test_schema_uri_mapping():
```
eslavich, line 7: I added a similar test in #14 (https://github.com/spacetelescope/rad/blob/bde8b1b43b98124ba880617c6cbc6a8605e26bac/tests/test_integration.py), which one do you want to keep?

PaulHuwe:
This has not been updated in years. Closing.

PR #16, opened 2021-04-20
Minor schema corrections

perrygreenfield:
A small fix was needed for the observation schema (name mismatch between the properties and the required properties and propertyOrder, a fix to the flat field schema, and adding a file to .gitignore


PaulHuwe (approved):
LGTM

eslavich (approved):
LGTM

PR #20, opened 2021-04-22
Improve flowStyle and propertyOrder tests, update ndarray validators

eslavich:
Having flowStyle in the wrong place causes asdf to raise an error, so I thought we'd better have a check for that.  I also updated the ndarray properties to use the special-purpose validators that asdf offers for those.


PaulHuwe (approved):
TGTM

PR #22, opened 2021-05-03
Simplify metaschema references

eslavich:
Now that https://github.com/asdf-format/asdf/pull/987 has been merged, we can simplify the local references in this project's metaschema.


PaulHuwe (approved):
Pound away!

PR #27, opened 2021-05-12
Update asdf requirement to >= 2.8.0

eslavich:
asdf 2.8.0 is released, so we can use a real requirement here now.  Huzzah!


PaulHuwe (approved):
Hashin!

Issue #32, opened 2021-05-22
Add dq_def File


PaulHuwe:
Needs to be implemented once table styles are decided upon.

PaulHuwe:
Depends on issue #24

PaulHuwe:
Implemented, but table issue still remains.

PaulHuwe:
Table issue resolved with PR [#48].

Issue #24, opened 2021-05-07
Decide On Table Types


PaulHuwe:
Decide on the table type support needed for RAD schema yamls (dq_def, group, ref/wfi_img_photom, etc.). Astropy tables aren't presently supported, but may be expected downstream, so some R&D and decision malign required.  

Astropy implementation example (from group):
    group:
      title: Group parameters table
      tag: tag:astropy.org:astropy/table/table-1.0.0
      colnames: [ integration_number, group_number, end_day, end_milliseconds, end_submilliseconds,
                  group_end_time, number_of_columns, number_of_rows, number_of_gaps,
                  completion_code_number, completion_code_text, bary_end_time, helio_end_time ]
      columns:
        - name: integration_number
          datatype: int16
        - name: group_number
          datatype: int16
        - name: end_day
          datatype: int16
        - name: end_milliseconds
          datatype: int32
        - name: end_submilliseconds
          datatype: int16
        - name: group_end_time
          datatype: [ ascii, 26 ]
        - name: number_of_columns
          datatype: int16
        - name: number_of_rows
          datatype: int16
        - name: number_of_gaps
          datatype: int16
        - name: completion_code_number
          datatype: int16
        - name: completion_code_text
          datatype: [ ascii, 36 ]
        - name: bary_end_time
          datatype: float64
        - name: helio_end_time
          datatype: float64

ddavis-stsci:
This issue is done.

PR #39, opened 2021-05-23
Added dq support.

PaulHuwe:
This PR is to add DQ definition schema.

Tests: Passes standard tests

Docs: Unneeded

Ticket: #32

### DIFF
--- a/src/rad/resources/manifests/datamodels-1.0.yaml
+++ b/src/rad/resources/manifests/datamodels-1.0.yaml
@@ -29,6 +29,12 @@ tags:
   title: Pixel area array
   description: |-
     The schema for the pixel area array
+# Helper Object Modules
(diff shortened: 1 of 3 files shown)


PaulHuwe:
This depends on Issue #24

src/rad/resources/manifests/datamodels-1.0.yaml
```diff
@@ -29,6 +29,12 @@ tags:
   title: Pixel area array
   description: |-
     The schema for the pixel area array
+# Helper Object Modules
```
perrygreenfield, line 32: Should this be considered a reference file?

ddavis-stsci:
In JWST they are reference files so I think for Roman they would be also.

Issue #33, opened 2021-05-22
Implement group support


PaulHuwe:
Needs to be implemented once table styles are decided upon.

PaulHuwe:
Depends on issue #24

nden:
There's no need for group extension support (like the one in jwst) for Roman.

PR #40, opened 2021-05-23
Added group schema.

PaulHuwe:
This PR is to add the group schema.

Tests: Passes standard tests

Docs: Unneeded

Ticket: [#33]

### DIFF
--- a/src/rad/resources/manifests/datamodels-1.0.yaml
+++ b/src/rad/resources/manifests/datamodels-1.0.yaml
@@ -29,6 +29,12 @@ tags:
   title: Pixel area array
   description: |-
     The schema for the pixel area array
+# Helper Object Modules
+- tag_uri: asdf://stsci.edu/datamodels/roman/tags/group-1.0.0
+  schema_uri: asdf://stsci.edu/datamodels/roman/schemas/group-1.0.0
+  title: Group parameter schema
+  description: |-
+    Group parameter schema
 # Metadata Modules
 - tag_uri: asdf://stsci.edu/datamodels/roman/tags/exposure-1.0.0
   schema_uri: asdf://stsci.edu/datamodels/roman/schemas/exposure-1.0.0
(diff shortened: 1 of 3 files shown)


PaulHuwe:
This depends on Issue #24

src/rad/resources/schemas/group-1.0.0.yaml
```diff
@@ -0,0 +1,48 @@
+%YAML 1.1
```
perrygreenfield, line 1: Let's defer this until Ed and I settle the table question. Hopefully very soon, but support in the library may depend on the solution. I doubt this is needed by subsequent steps or am I wrong?

PaulHuwe: DQ needs the table structure defined as well, so this needs to be resolved very soon.

perrygreenfield: Well, that is an issue for DQ. But is this particular table holding up development?

PaulHuwe: Consulted with RCAL team, and the group table is not holding up development at present, just the dq.

src/rad/resources/manifests/datamodels-1.0.yaml
```diff
@@ -29,6 +29,12 @@ tags:
   title: Pixel area array
   description: |-
     The schema for the pixel area array
+# Helper Object Modules
+- tag_uri: asdf://stsci.edu/datamodels/roman/tags/group-1.0.0
```
eslavich, line 33: Is group really intended to be a tagged object?  In the ASDF this will look something like this:

```yaml
group: !asdf://stsci.edu/datamodels/roman/tags/group-1.0.0
  group: !core/ndarray-1.0.0
    datatype: ...
```

(the double "group" is because the tagged object must be assigned to some property, but the schema for group also requires a property called "group")

src/rad/resources/schemas/group-1.0.0.yaml
perrygreenfield: After further discussions, we are at somewhat of a loss to understand why we need a group schema at all. It isn't clear yet that any JWST pipeline step that Roman would be using would ever use this information, and if it did, we likely could put this information into the data model without getting it from an ASDF file.

PaulHuwe:
We are dropping the group schema & datamodel as per the 06/22/21 Roman Development meeting.

Issue #34, opened 2021-05-22
Implement photom support


PaulHuwe:
Needs to be implemented once table styles are decided upon.

PaulHuwe:
Depends on issue #24

stscijgbot-rstdms:
_This issue is tracked on JIRA as [RAD-20](https://jira.stsci.edu/browse/RAD-20)._

stscijgbot-rstdms:
_Comment by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe) on [JIRA](https://jira.stsci.edu/browse/RAD-20?focusedCommentId=535090#comment-535090):_

This has been fully implemented, complete with newly required testing, PR comments addressed, PR approved and merged.

PR #41, opened 2021-05-23
Added photom support.

PaulHuwe:
This PR is to add photom.

Tests: Passes standard tests

Docs: Unneeded

Ticket: [#34]

### DIFF
--- a/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
@@ -14,7 +14,7 @@ properties:
     title: Photometric flux conversion factors table
     type: object
     patternProperties:
-      "^('F062'|'F087'|'F106'|'F129'|'W146'|'F158'|'F184'|'GRISM'|'PRISM'|'DARK'|'ENGINEERING')$":
+      "^(F062|F087|F106|F129|W146|F158|F184|GRISM|PRISM|DARK|ENGINEERING)$":
         type: object
         properties:
           photmjsr:
--- a/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
@@ -12,16 +12,18 @@ properties:
       - $ref: ref_common-1.0.0
   phot_table:
     title: Photometric flux conversion factors table
-    # Should this be an Astropy Table?
-    datatype:
-      - name: optical_element
-        datatype: [ascii, 12]
-      - name: photmjsr
-        title: surface brightness, in MJy/steradian
-        datatype: float32
-      - name: uncertainty
-        title: uncertainty of surface brightness, in MJy/steradian
-        datatype: float32
+    type: object
+    patternProperties:
+      "^('F062'|'F087'|'F106'|'F129'|'W146'|'F158'|'F184'|'GRISM'|'PRISM'|'DARK'|'ENGINEERING')$":
+        type: object
+        properties:
+          photmjsr:
+            title: surface brightness, in MJy/steradian
+            datatype: float32
+          uncertainty:
+            title: uncertainty of surface brightness, in MJy/steradian
+            datatype: float32
+    additionalProperties: false
 required: [meta, phot_table]
 flowStyle: block
 propertyOrder: [meta, phot_table]
--- a/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
@@ -23,6 +23,7 @@ properties:
           uncertainty:
             title: uncertainty of surface brightness, in MJy/steradian
             datatype: float32
+        required: [photmjsr, uncertainty]
     additionalProperties: false
 required: [meta, phot_table]
 flowStyle: block
--- a/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
@@ -1,0 +1,28 @@
+%YAML 1.1
+---
+$schema: asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.0.0
+id: asdf://stsci.edu/datamodels/roman/schemas/reference_files/wfi_img_photom-1.0.0
+
+title: WFI imaging photometric flux conversion data model
+
+type: object
+properties:
+  meta:
+    allOf:
+      - $ref: ref_common-1.0.0
+  phot_table:
+    title: Photometric flux conversion factors table
+    # Should this be an Astropy Table?
+    datatype:
+      - name: optical_element
+        datatype: [ascii, 12]
+      - name: photmjsr
+        title: surface brightness, in MJy/steradian
+        datatype: float32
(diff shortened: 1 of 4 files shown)


PaulHuwe:
This depends on Issue #24

src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
```diff
@@ -0,0 +1,28 @@
...
+title: WFI imaging photometric flux conversion data model
+
+type: object
+properties:
+  meta:
+    allOf:
+      - $ref: ref_common-1.0.0
+  phot_table:
+    title: Photometric flux conversion factors table
+    # Should this be an Astropy Table?
```
ddavis-stsci, line 15: I think this would be a natural place to use an astropy table. This is well isolated, the only other JWST code that uses it in in TSO mode and we don't need to worry about that.

ddavis-stsci (approved):
LGTM

src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
perrygreenfield: There are few enough optical elements that I wonder if an astropy table is overkill. I think a more natural organization is a dict where the keys are optical elements, and the contents of each are a dict with keys `photmjsr `and `uncertainty`. In other words, the properties are any one of the permitted optical elements, and the contents are another object with properties photmisr and uncertainty. I think JSONSchema provides a compact way of describing this, but I will look that up and add it (after having Ed grade my work).

```
type: object
patternProperties:
  "^('filter1'|'filter2'|'bozofilter3')$" :
    type: object
    properties:
      photmjsr:
        type: number
        title: blah, blah, blah megajanskies(?) per steradian
      uncertainty;
        type: number
        title: uncertainty of photmjsr
additionalProperties: false
```

src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
```diff
@@ -0,0 +1,28 @@
...
+    datatype:
```
eslavich, line 16: If we end up going with ndarray like this `datatype` validator suggests, we'll also need a `tag` validator

src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
```diff
@@ -0,0 +1,30 @@
...
+type: object
+properties:
+  meta:
+    allOf:
+      - $ref: ref_common-1.0.0
+  phot_table:
+    title: Photometric flux conversion factors table
+    type: object
+    patternProperties:
+      "^('F062'|'F087'|'F106'|'F129'|'W146'|'F158'|'F184'|'GRISM'|'PRISM'|'DARK'|'ENGINEERING')$":
```
eslavich, line 17: The single quotes here require that the value include literal single quotes.  Is that really what we want?  The ASDF file will have to look something like this:

```yaml
phot_table:
  "'F062'":
    photmjsr: ...
    uncertainty: ...
```

src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
```diff
@@ -0,0 +1,30 @@
...
+        type: object
+        properties:
+          photmjsr:
```
eslavich, line 20: Is this missing a "required" validator for photmjsr and uncertainty?

tests/test_schemas.py
```diff
@@ -14,6 +15,8 @@
     u for u in asdf.get_config().resource_manager
     if u.startswith(SCHEMA_URI_PREFIX) and u != METASCHEMA_URI
 ]
+WFI_OPTICAL_ELEMENTS = ["F062", "F087", "F106", "F129", "W146", "F158", "F184", "GRISM", "PRISM",
```
eslavich, line 18: Rather than maintain a source of truth here, I recommend declaring that the enum list in the wfi_mode-1.0.0 schema is canonical, and read the values from there.

tests/test_schemas.py
```diff
@@ -119,3 +122,29 @@ def callback(node):
             assert node["tag"] in valid_tag_uris
 
     asdf.treeutil.walk(schema, callback)
+
+
+# Confirm that WFI_OPTICAL_ELEMENTS matches the optical_element keyword attribute in wfi_mode.yaml
+def test_verify_optical_element_list(schema):
```
eslavich, line 128: If you take the suggestion up above on reading WFI_OPTICAL_ELEMENTS from wfi_mode-1.0.0, then this test won't be necessary.

tests/test_schemas.py
```diff
@@ -119,3 +122,29 @@ def callback(node):
...
+                assert node.get("type") == "object"
+                wfi_mode_oe_lst = list(node["properties"]["optical_element"]["enum"])
+                assert (set(wfi_mode_oe_lst) == set(WFI_OPTICAL_ELEMENTS))
+
+        asdf.treeutil.walk(schema, callback)
+
+
+# Confirm that the optical_element filter in wfi_img_photom.yml matches WFI_OPTICAL_ELEMENTS
+def test_matched_optical_element_entries(schema):
+    if (schema['id'].rsplit("/")[-1].split("-")[0] == "wfi_img_photom"):
```
eslavich, line 141: It takes some time to set up and tear down a test, and this will show up in the results as one passing test per schema.  I'd advise that we drop the "schema" fixture and instead look up the schema of interest by URI.

tests/test_schemas.py
```diff
@@ -119,3 +122,29 @@ def callback(node):
...
+        def callback(node):
+            if isinstance(node, Mapping) and "propertyOrder" in node:
+                assert node.get("type") == "object"
+                phot_table_keys = list(node["properties"]["phot_table"]["patternProperties"])
+                r = re.compile(phot_table_keys[0])
+                for element_str in WFI_OPTICAL_ELEMENTS:
+                    assert r.search("'" + element_str + "'")
+
+        asdf.treeutil.walk(schema, callback)
```
eslavich, line 150: I don't know that we need to walk the schema when we know the path already:

```python
phot_table_keys = list(schema["properties"]["phot_table"]["patternProperties"])
```

Might be less brittle that way, anyhow, since the callback will fail if we add any other object with a "propertyOrder" property.

tests/test_schemas.py
PaulHuwe: Done

tests/test_schemas.py
PaulHuwe: removed

src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
PaulHuwe: Good catch. Done.

tests/test_schemas.py
PaulHuwe: Simplified.

tests/test_schemas.py
PaulHuwe: Implemented.

src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
PaulHuwe: Fixed.

eslavich (approved):
LGTM

Issue #35, opened 2021-05-22
Add Support For Flat & Dark Reference Files


PaulHuwe:
This will also include the addition of ref_common. 

RAD-15

PaulHuwe:
Added.

Issue #25, opened 2021-05-07
Update WFI Data Files


PaulHuwe:
Update WFI data files to utilize helper data objects (like variance, etc.)

nden:
@PaulHuwe There's a PR linked. Can this be closed or is there more work to be done?

PaulHuwe:
> @PaulHuwe There's a PR linked. Can this be closed or is there more work to be done?

I think that is a mis-link. This issue has not been addressed, to my knowledge.

PR #42, opened 2021-05-23
Added dark and updated flat schemas.

PaulHuwe:
This PR is to add the dark schema and update the flat reference yaml.

Tests: Passes standard tests

Docs: Unneeded

Ticket: [#35]

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,8 @@
 
 - Added misc. required db keyword attributes. [JIRA RAD-7]
 
+- Added Dark schema and updated Flat schema. [#35]
+  
   
 0.1.0 (unreleased)
 ==================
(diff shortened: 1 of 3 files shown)


src/rad/resources/schemas/reference_files/flat-1.0.0.yaml
```diff
@@ -11,14 +11,17 @@ properties:
...
   data:
+    title: Flat data array
     tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
     datatype: float32
     ndim: 2
   dq:
+    title: Data quality array
     tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
-    datatype: uint32
+    datatype: uint16
```
ddavis-stsci, line 21: uint16 or 32? 
RAD still has this as uint32 in wfi-image-1.0.0
dq:
                datatype: uint32

ddavis-stsci (approved):
Other than DQ integer type it looks good.

perrygreenfield (approved):

src/rad/resources/schemas/reference_files/flat-1.0.0.yaml
PaulHuwe: This is to be tackled in Issue #25

Issue #38, opened 2021-05-23
Implement Variance Schema


PaulHuwe:

PaulHuwe:
Approved and merged.

PR #43, opened 2021-05-23
Initial ramp fitting support.

PaulHuwe:
This PR is to add ramp fitting support.

Tests: Fails as it depends on [#38]

Docs: Unneeded

Ticket: [#36]

PR #45, opened 2021-05-23
Added variance support.

PaulHuwe:
This PR is to add variance schema.

Tests: Passes standard tests

Docs: Unneeded

Ticket: [#38]

### DIFF
--- a/src/rad/resources/schemas/variance-1.0.0.yaml
+++ b/src/rad/resources/schemas/variance-1.0.0.yaml
@@ -1,0 +1,30 @@
+%YAML 1.1
+---
+$schema: asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.0.0
+id: asdf://stsci.edu/datamodels/roman/schemas/variance-1.0.0
+
+title: |
+  An object that is required to include the variance arrays.
+
+
+type: object
+properties:
+  var_poisson:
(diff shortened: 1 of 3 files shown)


src/rad/resources/schemas/variance-1.0.0.yaml
```diff
@@ -0,0 +1,30 @@
+%YAML 1.1
```
perrygreenfield, line 1: Will there be other variants? For example, is this what will be used for image products (3 dimensions for var_poisson or flat field?)

PaulHuwe: This will be the only variant for ELPP with the exception of the segment-specific flavour needed in rampfitoutput.

PaulHuwe: I have updated PR [#43] to remove this reference and install the segment-specific versions.

perrygreenfield (approved):
LGTM

src/rad/resources/schemas/variance-1.0.0.yaml
nden: Perhaps I misunderstand this but isn't the image variance, i.e. variance after ramp fitting a 2D array?

PR #48, opened 2021-06-03
Make dq_def a root level property

eslavich:
The existing schemas nest dq_def within the meta object, which is not where stcal expects to find it.


perrygreenfield (approved):
LGTM

PaulHuwe (approved):
TGTM

PR #49, opened 2021-06-03
Validate each reference's specific reftype

eslavich:
This prevents files from mixing and matching e.g. the FLAT reftype with the dark-1.0.0 tag.


PaulHuwe (approved):
LGTM

perrygreenfield (approved):
LGTM

PR #50, opened 2021-06-03
Add tag-and-release workflow

eslavich:
This adds a workflow that tags and releases a new version of the package.  I'm trying to make releases low-friction so that we can do them every day if we need to.

The workflow may have bugs, but I'm not sure how to test it until we actually want a release.


PaulHuwe (approved):
TGTM

PR #58, opened 2021-06-26
Fixed a few issues that cropped up from testing (Rampfitoutput & WFI_ing_photom).

PaulHuwe:
Ticket to fix a couple inaccuracies and omissions in rampfitoutput and wfi_img_photom.

### DIFF
--- a/src/rad/resources/schemas/rampfitoutput-1.0.0.yaml
+++ b/src/rad/resources/schemas/rampfitoutput-1.0.0.yaml
@@ -34,7 +34,7 @@ allOf:
     pedestal:
       title: Pedestal array
       tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
-      ndim: 2
+      ndim: 3
       datatype: float32
     weights:
       title: Weights for segment-specific fits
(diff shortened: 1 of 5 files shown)


ddavis-stsci (approved):
LGTM

src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
```diff
@@ -10,9 +10,13 @@ properties:
...
     allOf:
       - $ref: ref_common-1.0.0
+      - type: object
+        properties:
+          reftype:
+            type: string
+            enum: [PHOTOM]
   phot_table:
     title: Photometric flux conversion factors table
-    type: object
```
eslavich, line 15: Why is the type being removed?  I don't think it's valid to have `patternProperties` without `type: object`

src/rad/resources/schemas/rampfitoutput-1.0.0.yaml
```diff
@@ -34,7 +34,7 @@ allOf:
     pedestal:
```
eslavich, line 34: Should we change this schema URI and filename to `ramp_fit_output-1.0.0` so we end up with better consistency in capitalized words in roman_datamodels?

PaulHuwe: I can't think of a good not to. I will change the file_name.

src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
PaulHuwe: Because python throws errors about keys of type object.

eslavich (approved):
LGTM

PR #59, opened 2021-07-01
Remove basic from ref_common as well as reftype

### LINKED ISSUE #54: reference files should not require most of the information in basic-1.0.0
Most of the information in this schema is irrelevant for reference files, and can be very misleading. CRDS context, for example, should not be in the file since it may be used in several contexts. Likewise for calibration_software_version, crds_software_version, prd_software_version. Telescope is already required in another schema. If file date and such are needed, they should go in the ref_common schema and basic-1.0.0 removed from ref_common.

perrygreenfield:
Much of what is in basic is irrelevant to reference files and should not be required in reference files. This PR removes that $ref of basic, and transfers the presumably required attributes directly to ref_common (as I understand it, this includes filename, file_date, telescope, and origin; if not, please comment).

Also removed is reftype since each reference file schema should include this attribute with a single enum value.

This closes #54

### DIFF
--- a/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
@@ -31,5 +31,5 @@ allOf:
     origin:
       title: Organization responsible for creating file
       type: string
-  required: [reftype, author, description, pedigree, useafter, origin]
+  required: [reftype, author, description, pedigree, useafter, telescope, origin]
 ...
--- a/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
@@ -20,6 +20,10 @@ allOf:
     useafter:
       title: Use after date of the reference file
       tag: tag:stsci.edu:asdf/time/time-1.1.0
+    telescope:
+      title: Telescope data reference data is used to calibrate
+      type: string
+      enum: [ROMAN]
     filename:
       title: Name of the file
       type: string
--- a/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
@@ -6,14 +6,8 @@ id: asdf://stsci.edu/datamodels/roman/schemas/reference_files/ref_common-1.0.0
 title: Common reference metadata properties
 
 allOf:
-- $ref: ../basic-1.0.0
 - type: object
   properties:
-    reftype:
-      title: Reference file type
-      type: string
-      enum: [AREA, DARK, FLAT, GAIN, LINEARITY, MASK, PERSAT, PHOTOM, READNOISE, REFPIX,
-             SATURATION, SUPERBIAS, TRAPDENSITY, TRAPPARS]
     pedigree:
       title: The pedigree of the reference file
       type: string
@@ -26,5 +20,14 @@ allOf:
     useafter:
       title: Use after date of the reference file
       tag: tag:stsci.edu:asdf/time/time-1.1.0
-  required: [author, description, pedigree, useafter, reftype]
+    filename:
+      title: Name of the file
+      type: string
+    file_date:
+      title: Date this file was created (UTC)
+      tag: tag:stsci.edu:asdf/time/time-1.1.0
+    origin:
+      title: Organization responsible for creating file
+      type: string
+  required: [author, description, pedigree, useafter, filename, file_date, origin]
 ...
--- a/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
@@ -8,6 +8,9 @@ title: Common reference metadata properties
 allOf:
 - type: object
   properties:
+    reftype:
+      title: Reference File type
+      type: string
     pedigree:
       title: The pedigree of the reference file
       type: string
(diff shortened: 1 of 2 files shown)


src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
```diff
@@ -6,14 +6,8 @@ id: asdf://stsci.edu/datamodels/roman/schemas/reference_files/ref_common-1.0.0
...
 
 allOf:
-- $ref: ../basic-1.0.0
 - type: object
   properties:
-    reftype:
-      title: Reference file type
-      type: string
-      enum: [AREA, DARK, FLAT, GAIN, LINEARITY, MASK, PERSAT, PHOTOM, READNOISE, REFPIX,
-             SATURATION, SUPERBIAS, TRAPDENSITY, TRAPPARS]
```
PaulHuwe, line 16: Is this being removed due to duplication? Accordingly, would we then need to add reftype to all the individual reference models in their propertyOrder & required lists?

src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
```diff
@@ -26,5 +20,18 @@ allOf:
...
+    filename:
+      title: Name of the file
+      type: string
+    file_date:
+      title: Date this file was created (UTC)
+      tag: tag:stsci.edu:asdf/time/time-1.1.0
+    origin:
+      title: Organization responsible for creating file
+      type: string
+  required: [author, description, pedigree, useafter, filename, file_date, origin]
```
PaulHuwe, line 36: What about model_type?

PaulHuwe (changes requested):

src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
perrygreenfield: model_type was present for JWST since the FITS files don't have tags. The tag drives the model type in this case. It is redundant to the tag, and in any case, the tag is the primary indicator.

src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
perrygreenfield: I see that it isn't but I forget the subtleties of where you can put required when references are used. I'll let him comment on that. Maybe he can suggest the best solution.

src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
PaulHuwe: If the tag is driving the model, should model_type be removed from basic as well?

src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
eslavich: The `required` validator is independent of the individual property subschemas, so it should work fine to include reftype in the required list in this file.  But even so it might be worth keeping this much here:

```yaml
    reftype:
      title: Reference file type
      type: string
 ```
 
 mainly as documentation as to what reftype is supposed to be.  Without the enum list, so that doesn't need to be maintained.

File: src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml [hunk 3]
```diff
@@ -26,5 +20,18 @@ allOf:
     useafter:
       title: Use after date of the reference file
       tag: tag:stsci.edu:asdf/time/time-1.1.0
-  required: [author, description, pedigree, useafter, reftype]
+    telescope:
+      title: Telescope data reference data is used to calibrate
+      type: string
+      enum: [ROMAN]
+    filename:
```
eslavich, line 27: Does this need to be maintained by CRDS?  I mean, when CRDS receives a reference file submission and renames the file, should it also get into the file and update `meta.filename`?

perrygreenfield: I wondered about that as well, but when I asked @nden about this she said redcat required theses keywords (actually, I'm not exactly sure which of this set are required since I think I only asked specifically about file_date). I think she can answer this or tell us who can.

src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
perrygreenfield: I think that is a good solution. I'll put that in.

src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
eslavich: Matt says he's not sure where the requirement for meta.filename comes from and it sounds like it may not be necessary.  We have a meeting next week where I'll be able to ask more questions about it.

src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
PaulHuwe: This is resolved for me.

eslavich (approved):
LGTM!

src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
PaulHuwe: @perrygreenfield ?

PaulHuwe (approved):
LGTM, though the question about model_type is unanswered.

src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
perrygreenfield: It should. I can do that in a few minutes.

perrygreenfield: Remove it, that is.

PR #61, opened 2021-07-10
Updated DQ Arrays & Refout/Zeroframe cleaning.

### LINKED ISSUE #55: wfi_science_raw data arrays
Roman will not have a `zeroframe` and `refout`. Remove them from `wfi_science_raw`.

PaulHuwe:
…arrays.Closes #55 

Resolves RAD-24,22

Description:
Updated dq arrays to be of type uint32. Removed dq_def from being required (only needed if overriding). Removed zeroframe and refout arrays.

Checklist

[ x] Tests

[ ] Documentation

 Change log

[ ] Milestone

[ ] Label(s)

  (use "git restore --staged <file>..." to unstage)
	new file:   .idea/.gitignore
	new file:   .idea/inspectionProfiles/profiles_settings.xml
	new file:   .idea/modules.xml
	new file:   .idea/rad.iml
	new file:   .idea/vcs.xml


perrygreenfield:
Why is dq_def no longer required?

PaulHuwe:
> Why is dq_def no longer required?

It never should have been required. dq_def is an optional redefinition of dq flags.. if the reference file is using the default definitions, there is no need to override them (and thus, no need to waste space in files on them).

eslavich (approved):
LGTM.  We'll need a change in stcal if we want to support optional dq_def, currently the dynamic_mask function assumes that dq_def exists:

https://github.com/spacetelescope/stcal/blob/main/src/stcal/dynamicdq.py#L27

Can you file an issue over there if you and @perrygreenfield agree to make it optional?

PaulHuwe:
> LGTM. We'll need a change in stcal if we want to support optional dq_def, currently the dynamic_mask function assumes that dq_def exists:
> 
> https://github.com/spacetelescope/stcal/blob/main/src/stcal/dynamicdq.py#L27
> 
> Can you file an issue over there if you and @perrygreenfield agree to make it optional?

Yeah, or I can put it in rdm (explicitly set dq_def = None, if it doesn't exist).

perrygreenfield:
I haven't thought this through completely, but I wonder if the details of dq_def ought to be documented in the reference file schemas. If the are to be added, then we probably should be explicit about that. And if not, we probably should leave them out entirely rather than make them optional. I suspect that the pipeline code will depend on their presence and as such, versioning them based on that may be useful, as well as their specific bit plane names. Comments @nden and @eslavich?

PaulHuwe:
Could a solution (to both) be to have dq_def be required, but allow it to be empty? That way we don't need to worry about an additional check, or having to add a lot of versioning machinery for them?

perrygreenfield:
I've been trying to find out more about past usage. In general, any dq_def column definitions that don't correspond to the standard set defined for data files are never used by JWST pipelines. The instrument teams do add them, apparently as some sort of aid in making the reference files, and they get passed along. Personally, I don't think those sorts of DQ bits should be part of delivered reference files. The main justification for the other use is only selecting relevant DQ items thus saving space in the reference file, i.e., only using 8 or 16 bits instead of 32. To this there are two comments. If space is really an issue, lossless compression can be used for arrays, in which case whether they are 8 bits or 32 bits is essentially irrelevant for the space used. I tend to avoid recommending compression for archival purposes, but on the other hand, these are reference files, and I suspect their archival use fades fairly fast with time. Is suspect the standard compression algorithms will greatly outlast the usefulness of the reference file DQ array.

Finally @jdavies-st believes these dq_def items have been much more of a pain leading to many bugs and problems for JWST and he does not recommend their use, and @hbushouse also feels that they weren't worth the trouble. So can we consider just getting rid of dq_def for Roman?

ddavis-stsci:
I agree with Perry. Unless we have a clear use case for this feature I would leave it out. If the instrument teams need them we can add them back.

PaulHuwe:
I am fine for removing dq_def from Roman entirely. @ddavis-stsci @nden Who needs to make the call on this, so I can remove it?

PaulHuwe:
This PR is paused until INS makes a determination on dynamic DQ support.

perrygreenfield:
I'm not sure it needs to be paused. If it is only needed for their purposes, then it need not be included in the schema. Unless it is the size issue they are concerned about. If that is the issue, are they aware that we can compress the DQ array so that size is not an issue?

PaulHuwe:
This PR is unblocked, and work is resuming.

PaulHuwe:
DQ Def code removed.

eslavich (approved):
This LGTM, but it did before too, so take that with a grain of salt :)

perrygreenfield (approved):
LGTM

ddavis-stsci (approved):
LGTM

PR #65, opened 2021-08-09
Add enumeration for pedigree values

nden:
`PEDIGREE` is defined only as type string, however CRDs validates it against the following values:
`[INFLIGHT, GROUND, MODEL, DUMMY, SIMULATION]` and so files with arbitrary string values fail validation in CRDS.
This PR adds the enumeration to the schema.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 0.6.0 (unreleased)
 ==================
 
+- Added enumeration for ``meta.pedigree``. [#65]
+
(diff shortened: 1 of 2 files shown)


PaulHuwe (approved):
LGTM

nden:
Followed Ed's suggestion and removed `INFLIGHT` as one of the options. The reason is it can have date range associated with it. We'll need to implement it in the future but not for a while.

PR #79, opened 2021-09-17
RAD-32 Added nonlinearity support.

PaulHuwe:
Added support for linearity correction reference files.


gardner358 (approved):

ddavis-stsci (approved):
Looks good

PR #80, opened 2021-09-18
RAD-29 Adding support for pixel area reference files.

PaulHuwe:
Adding support for pixel area reference files.


gardner358 (approved):

ddavis-stsci (approved):
LGTM

PR #83, opened 2021-09-20
some attributes should not be required

nden:
This PR removes `pixelarea` and `var_flat` from the list of required attributes in the `wfi_image` schema.
The `wfi_image` schema is used for the first time with the output of ramp fitting. `pixelarea` and `var_flat` are not available at that time and the validation of the output of ramp_fitting fails.

### DIFF
--- a/src/rad/resources/schemas/pixelarea-1.0.0.yaml
+++ b/src/rad/resources/schemas/pixelarea-1.0.0.yaml
@@ -15,5 +15,4 @@ properties:
     ndim: 2
 propertyOrder: [area]
 flowStyle: block
-required: [area]
 ...
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@
 
 - Added pixel area reference file support. [#80]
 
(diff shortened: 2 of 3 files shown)


eslavich (approved):
LGTM

src/rad/resources/schemas/pixelarea-1.0.0.yaml
```diff
@@ -15,5 +15,4 @@ properties:
     ndim: 2
 propertyOrder: [area]
 flowStyle: block
-required: [area]
```
PaulHuwe, line 18: Do we need to remove the requirement here? Isn't its absence in wfi_image-yaml sufficient? @eslavich

eslavich: wfi_image-1.0.0 has a $ref to pixelarea-1.0.0, so we also need to remove the "required" here.

PaulHuwe: Gotcha. I will take my further questions offline.

PaulHuwe (approved):

PR #86, opened 2021-09-24
Pedestal array needs to be 2D

nden:
The `pedestal` array in the optional output from ramp fitting, `ramp_fit_output`, represents the signal at zero exposure time for each pixel. Since Roman has one integration only I believe this array needs to be 2 dimensional.


ddavis-stsci (approved):
I agree, it should be 2-d not 3-d.

Issue #87, opened 2021-09-24
Add Mode To Reference Schemas


stscijgbot-rstdms:
_Issue [RAD-38](https://jira.stsci.edu/browse/RAD-38) was created on JIRA by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe):_

The following reference files are viewing mode dependent and thus require the exposure type as a CRDS selector.

The schemas (list from INS):

Dark

Readnoise

stscijgbot-rstdms:
_Comment by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe) on [JIRA](https://jira.stsci.edu/browse/RAD-38?focusedCommentId=551154#comment-551154):_

This is in PR:
#90

stscijgbot-rstdms:
_Comment by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe) on [JIRA](https://jira.stsci.edu/browse/RAD-38?focusedCommentId=551658#comment-551658):_

I refactored the code to address the comments, remove any concerns of unwanted database keywords in reference files, and allow for flexibility with reference exposure keywords. It is still in review.

stscijgbot-rstdms:
_Comment by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe) on [JIRA](https://jira.stsci.edu/browse/RAD-38?focusedCommentId=551659#comment-551659):_

The ticket for the associated RDM work is RCAL-225:
<https://jira.stsci.edu/browse/RCAL-225>

nden:
What is the motivation for this change? Are these reference files going to be dependent on exposure type and so `meta.exposure.type` will be a CRDS selector? If so please make this clear in the ticket.

stscijgbot-rstdms:
_Comment by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe) on [JIRA](https://jira.stsci.edu/browse/RAD-38?focusedCommentId=551698#comment-551698):_

I have updated the ticket, and will update RAD-39 as well.

PR #90, opened 2021-09-27
RAD-38 Mode keyword for Dark and Readnoise

PaulHuwe:
Added mode keyword to dark and readnoise.

### DIFF
--- a/src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
@@ -15,6 +15,11 @@ properties:
           reftype:
             type: string
             enum: [DARK]
+      - type: object
+        properties:
+          mode:
+            $ref: asdf://stsci.edu/datamodels/roman/schemas/exposure-1.0.0#properties/type
+        required: [mode]
   data:
     title: Dark current array
     tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
--- a/src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
@@ -13,13 +13,8 @@ properties:
       - type: object
         properties:
           reftype:
-            type: string
             enum: [DARK]
-      - type: object
-        properties:
-          mode:
-            $ref: asdf://stsci.edu/datamodels/roman/schemas/exposure-1.0.0#properties/type
-        required: [mode]
+      - $ref: ref_exposure_type-1.0.0
   data:
     title: Dark current array
     tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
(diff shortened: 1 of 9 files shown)


PaulHuwe:
I implemented this by importing the type keyword from exposure. Exposure/type contains both sdf & archive_catalog. I am wondering if this is an issue or if having those is irrelevant.

If it is an issue, I can just make a new file just for reference file modes.

src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
```diff
@@ -15,6 +15,11 @@ properties:
           reftype:
             type: string
             enum: [DARK]
+      - type: object
+        properties:
+          mode:
+            $ref: asdf://stsci.edu/datamodels/roman/schemas/exposure-1.0.0#properties/type
```
gardner358, line 21: It seems to me that you are including exposure's type keyword.  This points to an archive_catalog entry.  I don't think it will be a problem but I can't be sure.  Same for READNOISE.

gardner358 (approved):

src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
PaulHuwe: Yeah, this is what my question above is about. I did it this way to avoid duplicating code and keeping multiple enum lists for exposure mode/type up to date, but if there are issues, we can do that.

gardner358: I think you should leave it.  I'm hoping it won't affect anything.

PaulHuwe:
Issue #87

src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
eslavich: I'd recommend linking directly to the enum list, instead of the entire `type` property.  That will avoid bringing in the `sdf` and `archive_catalog` objects and also give us the opportunity to include a `title` here that is more relevant to the dark file.

eslavich: Incidentally, why is this property at meta.mode instead of meta.exposure.type?  Would consistency make these easier to work with?

src/rad/resources/schemas/reference_files/readnoise-1.0.0.yaml
```diff
@@ -15,6 +15,11 @@ properties:
           reftype:
             type: string
             enum: [READNOISE]
+      - type: object
```
eslavich, line 18: Since this block is used in multiple files, it might be worth creating a small subschema that can be ref'ed in here with this content.

PaulHuwe: Reftype already is imported.. this is just setting its value.

PaulHuwe:
I refactored the code to address the comments, remove any concerns of unwanted database keywords in reference files, and allow for flexibility with reference exposure keywords.

gardner358 (approved):

ddavis-stsci (approved):
LGTM

PR #97, opened 2021-11-15
RAD-42 Change bit fields to nchar(1)

gardner358:
Changing archive_catalog bit datatype fields to nchar(1).


PaulHuwe (changes requested):
Things look fine, but I want to do an RDM test before approving. (Just a sanity check.)

PaulHuwe (approved):
RDM & RCAL build with no issue with these changes.

PR #101, opened 2021-12-14
Test with Python 3.10

nden:
@eslavich Any idea what causes the error?

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,6 +5,8 @@ envlist= py38,style,bandit
 extras= test
 commands=
     pytest
+deps=
+    py310: git+https://github.com/pytest-dev/pytest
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -14,6 +14,7 @@ commands=
 [testenv:twine]
 deps=
     twine
+    py310: git+https://github.com/pytest-dev/pytest
```
eslavich, line 17: This should be in the `[testenv]` section and not the `[testenv:twine]` section, I think that's why the Python 3.10 job is still failing.

nden: yes it should 🤦

nden:
@eslavich ready for review

PR #103, opened 2022-01-18
RAD-45 Add Guide window modes to guide star schema

ddavis-stsci:
Added guide window mode data. I've left the CGI off the mode list for now since it is not clear what we would do with it in that attribute.


PaulHuwe (approved):
LGTM

gardner358 (approved):

PR #104, opened 2022-01-19
RAD-10: Add attribute for compression in originating packets

### LINKED ISSUE #51: Add attribute for compression in originating packets
_Issue [RAD-10](https://jira.stsci.edu/browse/RAD-10) was created on JIRA by [Megan Sosey](https://jira.stsci.edu/secure/ViewProfile.jspa?name=sosey):_

The WFI packets will contain the compression flag that’s appropriate for the data, those are part of the L0 data, and not all L0 data will be compressed.

DMS will use the decompression software to decompress as appropriate, and the level 1 file we create will contain uncompressed data. 

 We should carry through some metadata into the L1 file that says whether the packets that were used to construct it arrived compressed or not, in case it’s useful for downstream error analysis – for the case that folks looking at the level 1+ data are trying to diagnose a general issue. I would expect errors with the actual decompression to be addressed immediately.

PaulHuwe:
Added boolean level0_compressed attribute keyword to exposure group to indicate if the level 0 data was compressed.

Closes #51
Resolves  [RAD-10](https://jira.stsci.edu/browse/RAD-10)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@
 
 - Set all calsteps to required. [#102]
 
+- Added boolean level0_compressed attribute keyword to exposure group to indicate if the level 0 data was compressed. [#104]
+
 0.8.0 (2021-11-22)
 ==================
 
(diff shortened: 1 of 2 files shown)


gardner358 (approved):

ddavis-stsci (approved):
LGTM

PR #105, opened 2022-01-19
RAD-44: Add p_keywords to readnoise and dark 

PaulHuwe:
Add p_exptype to exposure group for reference files (dark & readnoise) to enable automatic rmap generation. Added test to ensure that the p_exptype expression matched the exposure/type enum list.

### DIFF
--- a/src/rad/resources/schemas/reference_files/ref_exposure_type-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/ref_exposure_type-1.0.0.yaml
@@ -14,6 +14,12 @@ properties:
         allOf:
           - $ref: ../exposure_type-1.0.0
           - title: Type of data in the exposure (viewing mode)
-    required: [type]
+      p_exptype:
+        title: Applicable EXP_TYPE.
+          type: string
+          pattern: "\
+            ^((WFI_IMAGE|WFI_GRISM|WFI_PRISM|WFI_DARK|WFI_FLAT|\
+            WFI_WFSC)\\s*\\|\\s*)+$"
+    required: [type,p_exptype]
 required: [exposure]
 ...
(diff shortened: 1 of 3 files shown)


eslavich (approved):
LGTM!  Thanks for the test!

src/rad/resources/schemas/reference_files/ref_exposure_type-1.0.0.yaml
```diff
@@ -14,6 +14,10 @@ properties:
         allOf:
           - $ref: ../exposure_type-1.0.0
           - title: Type of data in the exposure (viewing mode)
-    required: [type]
+      p_exptype:
+        title: Applicable exposure type.
+        type: string
+        pattern: "^((WFI_IMAGE|WFI_GRISM|WFI_PRISM|WFI_DARK|WFI_FLAT|WFI_WFSC)\\s*\\|\\s*)+$"
+    required: [type,p_exptype]
```
ddavis-stsci, line 21: Maybe I just don't understand but why do we have WFI_DARK and WFI_FLAT here? I thought this was so that items like WFI_PRISM & WFI_GRISM can use the same ref files. So P_EXPTYPE in the reference file would refer to WFI_IMAGE|WFI_PRISM|WFI_GRISM
I'm not the expert here but if Ed says it works...

ddavis-stsci (approved):
If Ed says OK...

src/rad/resources/schemas/reference_files/ref_exposure_type-1.0.0.yaml
PaulHuwe: INS confirmed the list for this, so I kept it as is.. I didn't ask beyond that.

PR #115, opened 2022-02-11
Remove 'photometry' from 'required' attributes in 'common'

nden:
The `photometry` section is updated in the image file, in the last step of ELPP. 
Having it as a required property adds it to L1 files.
This PR makes it possible not to include it in L1 files. 
The alternative is to have it with dummy values.


ddavis-stsci (approved):
Since the photometry keywords only make sense for calibrated data I'm all for removing them from L1 files.

PaulHuwe (commented):
LGTM

PR #117, opened 2022-02-14
RAD-49 Updated dark schema to include group keywords from exposure.

PaulHuwe:
Updated dark schema to include group keywords from exposure.

### DIFF
--- a/src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
@@ -21,7 +21,20 @@ properties:
                 title: Identifier for the multi-accumulation table used
                 type: string
             required: [ma_table_name]
-        required: [observation]
+          exposure:
+            type: object
+            properties:
+              ngroups:
+                title: Number of groups in integration
+                type: integer
+              nframes:
+                title: Number of frames per group
+                type: integer
+              groupgap:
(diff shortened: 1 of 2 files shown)


gardner358 (approved):

src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
```diff
@@ -21,7 +21,20 @@ properties:
...
+            properties:
+              ngroups:
+                title: Number of groups in integration
+                type: integer
+              nframes:
+                title: Number of frames per group
+                type: integer
+              groupgap:
+                title: Number of frames dropped between groups
+                type: integer
```
eslavich, line 35: I don't know if you want to get into this, but there are validators available that constrain numbers to a range, see:

https://json-schema.org/understanding-json-schema/reference/numeric.html#range

PaulHuwe: This is a great point.. but I think one better saved for going through and tightening up everything (allowed ranges here, reasonable default in RDM, etc). @ddavis-stsci @nden If y'all agree, I can make a ticket for these.

ddavis-stsci: I don't see a problem with that. The RTB team should be involved when deciding valid ranges.

PR #118, opened 2022-02-14
tweaks to schemas

nden:
Changes:

- remove `photometry` from `common`
- change the datatype of `ramp.data` to float32
- change data types of attributes in pixelarea to `Quantity`
- remove `pixelarea` from Level2 schemas. The latest photometry algorithm specifies it won't be attached to files.

### DIFF
--- a/src/rad/resources/schemas/photometry-1.0.0.yaml
+++ b/src/rad/resources/schemas/photometry-1.0.0.yaml
@@ -54,9 +54,6 @@ properties:
     archive_catalog:
       datatype: float
       destination: [ScienceCommon.conversion_microjanskys_uncertainty]
-propertyOrder: [conversion_microjanskys, conversion_megajanskys,
-                pixelarea_steradians, pixelarea_arcsecsq,
-                conversion_megajanskys_uncertainty, conversion_microjanskys_uncertainty]
 flowStyle: block
 required: [conversion_microjanskys, conversion_megajanskys,
            pixelarea_steradians, pixelarea_arcsecsq,
--- a/src/rad/resources/schemas/photometry-1.0.0.yaml
+++ b/src/rad/resources/schemas/photometry-1.0.0.yaml
@@ -9,7 +9,7 @@ properties:
   conversion_megajanskys:
     title: Flux density (MJy/steradian) producing 1 cps
     anyOf:
-      - $ref: tag:stsci.edu:asdf/unit/quantity-1.1.0
+      - tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
       - type: "null"
     archive_catalog:
       datatype: float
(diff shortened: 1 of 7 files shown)


src/rad/resources/schemas/wfi_image-1.0.0.yaml
```diff
@@ -7,12 +7,12 @@ title: |
   The schema for WFI Level 2 images.
 
 allOf:
-  - $ref: pixelarea-1.0.0
   - type: object
     properties:
       meta:
         allOf:
           - $ref: common-1.0.0
+          - $ref: photometry-1.0.0
```
eslavich, line 15: Consider replacing this with a tag validator since it is a tagged object:

```yaml
tag: asdf://stsci.edu/datamodels/roman/tags/photometry-1.0.0
```

(common-1.0.0 is just a schema so we have to $ref it here)

src/rad/resources/schemas/reference_files/pixelarea-1.0.0.yaml
```diff
@@ -20,10 +20,14 @@ properties:
             properties:
               pixelarea_steradians:
                 title: Nominal pixel area in steradians
-                type: number
+                anyOf:
+                  - $ref: tag:stsci.edu:asdf/unit/quantity-1.1.0
```
eslavich, line 24: This should be `tag:` instead of `$ref`

src/rad/resources/schemas/reference_files/pixelarea-1.0.0.yaml
```diff
@@ -20,10 +20,14 @@ properties:
...
+                  - type: "null"
               pixelarea_arcsecsq:
                 title: Nominal pixel area in arcsec^2
-                type: number
+                anyOf:
+                  - $ref: tag:stsci.edu:asdf/unit/quantity-1.1.0
```
eslavich, line 29: Also `tag:` instead of `$ref`

src/rad/resources/schemas/photometry-1.0.0.yaml
```diff
@@ -54,9 +54,6 @@ properties:
     archive_catalog:
       datatype: float
       destination: [ScienceCommon.conversion_microjanskys_uncertainty]
-propertyOrder: [conversion_microjanskys, conversion_megajanskys,
```
eslavich, line 57: IIRC `propertyOrder` will no longer cause a problem if the `$ref` to this schema is changed to a `tag:` validator instead.

nden: Oh, I hope that's what's causing the problem

nden: OK, this worked. So the cryptic `dict has no attribute propertyOrder` error was triggered because it was declared a `ref` instead of `tag`. Makes sense now.

eslavich (approved):
LGTM

nden:
Thanks @eslavich for the help!

PR #119, opened 2022-02-17
RAD-54: Add Instrument and Detector to Reference Schemas

PaulHuwe:
Moved detector list to new file for importing to both data and reference schemas.

### DIFF
--- a/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
@@ -34,17 +34,12 @@ allOf:
     instrument:
       type: object
       properties:
-        telescope:
-          type: object
-          properties:
-            name:
-              title: Instrument used to acquire the data
-              type: string
-              enum: [WFI]
-            detector:
-              allOf:
-                - $ref: ../wfi_detector-1.0.0
-          required: [name, detector]
-      required: [telescope]
+        name:
+          title: Instrument used to acquire the data
+          type: string
+          enum: [WFI]
+        detector:
+          $ref: ../wfi_detector-1.0.0
+      required: [name, detector]
   required: [reftype, author, description, pedigree, useafter, telescope, origin, instrument]
 ...
--- a/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
@@ -28,9 +28,23 @@ allOf:
       title: Telescope data reference data is used to calibrate
       type: string
       enum: [ROMAN]
-      type: string
     origin:
       title: Organization responsible for creating file
       type: string
-  required: [reftype, author, description, pedigree, useafter, telescope, origin]
+    instrument:
+      type: object
+      properties:
+        telescope:
+          type: object
+          properties:
+            name:
+              title: Instrument used to acquire the data
+              type: string
+              enum: [WFI]
+            detector:
+              allOf:
+                - $ref: ../wfi_detector-1.0.0
+          required: [name, detector]
+      required: [telescope]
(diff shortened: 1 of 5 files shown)


src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
```diff
@@ -28,9 +28,23 @@ allOf:
...
       enum: [ROMAN]
-      type: string
     origin:
       title: Organization responsible for creating file
       type: string
-  required: [reftype, author, description, pedigree, useafter, telescope, origin]
+    instrument:
+      type: object
+      properties:
+        telescope:
```
gardner358, line 37: It seems you have telescope here  but it's still listed above.

PaulHuwe: I was duplicating the structure in the data side, and yeah, I agree.. going to remove a layer.

src/rad/resources/schemas/reference_files/ref_common-1.0.0.yaml
```diff
@@ -28,9 +28,23 @@ allOf:
...
+          type: object
+          properties:
+            name:
+              title: Instrument used to acquire the data
+              type: string
+              enum: [WFI]
+            detector:
+              allOf:
```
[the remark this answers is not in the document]
PaulHuwe: Removed

src/rad/resources/schemas/wfi_mode-1.0.0.yaml
```diff
@@ -20,17 +20,15 @@ properties:
...
-    archive_catalog:
-      datatype: nvarchar(10)
-      destination: [ScienceCommon.detector]
+    allOf:
+      - $ref: wfi_detector-1.0.0
+      - sdf:
+          special_processing: VALUE_REQUIRED
+          source:
+            origin: TBD
+      - archive_catalog:
```
eslavich, line 29: Would it be better not to place this into a separate object?  Everywhere else we seem to have `sdf` and `archive_catalog` together.  I just mean remove the `-` on this line.

PaulHuwe: I don't understand? These are properties of detector, which is importing an enum list (similar to optical_element directly below it).

eslavich: It may be clearer if I express it in JSON.  This is what you have:

```json
"detector": {
  "allOf": [
    {
      "$ref": "wfi_detector-1.0.0"
    },
    {
      "sdf": {
        "special_processing": "VALUE_REQUIRED",
        "source": {"origin": "TBD"}
      }
    },
    {
      "archive_catalog": {
         "datatype": "nvarchar(10)",
         "destination": ["ScienceCommon.detector"]
      }
    }
  ]
}
```

and I'm suggesting that it's more in line with the rest of the repo to express it like this:

```json
"detector": {
  "allOf": [
    {
      "$ref": "wfi_detector-1.0.0"
    },
    {
      "sdf": {
        "special_processing": "VALUE_REQUIRED",
        "source": {"origin": "TBD"}
      },
      "archive_catalog": {
         "datatype": "nvarchar(10)",
         "destination": ["ScienceCommon.detector"]
      }
    }
  ]
}
```

If those were validators it wouldn't matter, but it sounds like other code is going to be crawling the schema looking for the `sdf` and `archive_catalog` properties and may not expect them to be in different objects?

PaulHuwe: Thanks.. I fixed this as well as a few other places that had that formatting as well.

cshanahan1 (approved):
I can't find anything worth a suggestion, so I will give this a LGTM

eslavich (approved):
LGTM2

PR #128, opened 2022-03-16
RAD-59: Remove exptype and p_keywords from Distortion Model

PaulHuwe:
Removed exptype and p_keywords from Distortion schema.


ddavis-stsci (approved):
Looks good.

gardner358 (approved):

PR #134, opened 2022-04-02
RAD-57: Add ma_table_number to the schema

PaulHuwe:
Added ma_table_number to observation and dark schemas.

### DIFF
--- a/src/rad/resources/schemas/observation-1.0.0.yaml
+++ b/src/rad/resources/schemas/observation-1.0.0.yaml
@@ -195,6 +195,16 @@ properties:
     archive_catalog:
       datatype: nvarchar(50)
       destination: [ScienceCommon.ma_table_name]
+  ma_table_number:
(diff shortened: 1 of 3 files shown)


src/rad/resources/schemas/observation-1.0.0.yaml
```diff
@@ -195,6 +195,16 @@ properties:
...
+  ma_table_number:
+    title: Number of the multi-accumulation table used
+    type: integer
+    sdf:
+      special_processing: VALUE_REQUIRED
+      source:
+        origin: TBD
+    archive_catalog:
+      datatype: smallint
+      destination: [ScienceCommon.execution_plan]
```
gardner358, line 207: Why is the field name execution_plan?

PaulHuwe: Good catch! Fixed.

PR #139, opened 2022-04-12
Rad 51 Update the attribute descriptions in the exposure-1.0.0 schema

ddavis-stsci:
Updated the attribute descriptions and reviewed the use of UTC. All times that are stored in non-UTC times need the other system. The MJD dates are used for multi-mission searches & the TDB times cannot be expressed in UTC in a meaningful why.

### DIFF
--- a/src/rad/resources/schemas/exposure-1.0.0.yaml
+++ b/src/rad/resources/schemas/exposure-1.0.0.yaml
@@ -33,8 +33,9 @@ properties:
 
   start_time:
     title: UTC exposure start time
-    description: This is a python date-time object that records the
-         time at the start of the exposure in UTC.
+    description: |
+        This is a python date-time object that records the
+        time at the start of the exposure in UTC.
     tag: tag:stsci.edu:asdf/time/time-1.1.0
     sdf:
       special_processing: VALUE_REQUIRED
@@ -45,8 +46,9 @@ properties:
       destination: [ScienceCommon.exposure_start_time]
   mid_time:
     title: UTC exposure mid time
-    description: This is a python date-time object that records the
-         time at the middle of the exposure in UTC.
+    description: |
+        This is a python date-time object that records the
+        time at the middle of the exposure in UTC.
     tag: tag:stsci.edu:asdf/time/time-1.1.0
     sdf:
       special_processing: VALUE_REQUIRED
@@ -57,8 +59,9 @@ properties:
       destination: [ScienceCommon.exposure_mid_time]
   end_time:
     title: UTC exposure end time
-    description: This is a python date-time object that records the
-         time at the end of the exposure in UTC.
+    description: |
+        This is a python date-time object that records the
+        time at the end of the exposure in UTC.
     tag: tag:stsci.edu:asdf/time/time-1.1.0
     sdf:
       special_processing: VALUE_REQUIRED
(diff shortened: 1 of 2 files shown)


src/rad/resources/schemas/exposure-1.0.0.yaml
```diff
@@ -122,6 +150,9 @@ properties:
       destination: [ScienceCommon.exposure_end_time_tdb]
   ngroups:
     title: Number of groups in integration
+    description: This is the number of resultant frames in the exposure
+         that are transmitted to the ground. The WFI data always has the
+         number on integrations=1.
```
PaulHuwe, line 155:
```suggestion
         number of integrations=1.
```

src/rad/resources/schemas/exposure-1.0.0.yaml
```diff
@@ -172,7 +207,8 @@ properties:
       destination: [ScienceCommon.exposure_gain_factor]
   integration_time:
     title: "[s] Effective integration time"
-    type: number
+    description: The effective time that the sensor has been exposed to the sky. 
+     type: number
```
PaulHuwe, line 211:
```suggestion
    type: number
```

src/rad/resources/schemas/exposure-1.0.0.yaml
```diff
@@ -182,7 +218,9 @@ properties:
       destination: [ScienceCommon.exposure_integration_time]
   elapsed_exposure_time:
     title: "[s] Total elapsed exposure time"
-    type: number
+    description:  The time between the start of the first Reset/Read Science Frame of an Exposure
+            and the completion of the final Read Only Science Frame of that Exposure.
+     type: number
```
PaulHuwe, line 223:
```suggestion
    type: number
```

src/rad/resources/schemas/exposure-1.0.0.yaml
```diff
@@ -222,7 +265,9 @@ properties:
       destination: [ScienceCommon.exposure_frame_time]
   group_time:
     title: "[s] Time between groups"
-    type: number
+     description: The time that is the sum of the reads that are used to construct a resultant.
+             This will depend on the MA table being used. 
+   type: number
```
PaulHuwe, line 270:
```suggestion
     type: number
```

PaulHuwe (commented):
One small typo and some weird spacing issues (that may be causing the test failures?)

src/rad/resources/schemas/exposure-1.0.0.yaml
```diff
@@ -32,6 +33,8 @@ properties:
 
   start_time:
     title: UTC exposure start time
+    description: This is a python date-time object that records the
+         time at the start of the exposure in UTC.
```
nden, line 37: For multiline strings use `|`, e.g.
```
description:  |
  This is a python date-time object that records the
  time at the start of the exposure in UTC.
```

src/rad/resources/schemas/exposure-1.0.0.yaml
ddavis-stsci: Good catch: fixed

src/rad/resources/schemas/exposure-1.0.0.yaml
ddavis-stsci: Thanks All fixed. And it highlighted the fact my linter was broken.

src/rad/resources/schemas/exposure-1.0.0.yaml
ddavis-stsci: All multi-line strings now use |
I also added descriptions for ma_table_name and ma_table_number which I missed.

PaulHuwe (approved):
LGTM

nden (approved):
Nice!

PR #142, opened 2022-04-22
RAD-52 After review remove observation.start_time and end_time

ddavis-stsci:
After review with INS it was decided to remove start_time and end_time from the observation schema since no use cases could be found and the origin of these times cannot be determined.


nden (approved):

PaulHuwe (approved):
LGTM

PR #149, opened 2022-06-08
RAD-65: Update error array in ramp-1.0.0.yaml

PaulHuwe:
Changed the dimensionality of the err variable in ramp.


ddavis-stsci (approved):
LGTM

PR #157, opened 2022-07-27
fix ref_file schema

cshanahan1:
Fix indentation issue, and change types from 'object' to string - reference file names don't need to be in a dictionary since their only property is their name.


nden (approved):
LGTM

nden:
needs a rebase

PR #153, opened 2022-07-22
RAD-71 Move archive information to be in tags

WilliamJamieson:
Moves archive related information directly into a tagged object for easier retrieval by ASDF.

PR #158, opened 2022-07-28
Keep "telescope" attribute in ref_common of type string 

nden:
to avoid regenration of reference files
Follow up on #153 - reverted `ref_common.telescope` back to type `string` to avoid a last minute need to generate all reference files again.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,9 +7,14 @@
 
 - Create docs for RTD. [#151]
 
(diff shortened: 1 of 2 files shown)


perrygreenfield (approved):
LGTM

WilliamJamieson:
This will cause my changes to roman_datamodels to start failing the unit tests.

To avoid this you may want to put an anyOf or oneOf statement listing both this change and the tag version.

PR #161, opened 2022-08-23
Fix `ref_file` and and archive info to it

WilliamJamieson:
This PR accomplishes two things:

1. Fixes the indentation of the ref_file.
2. Adds archive information to the CRDS keywords in the file.

Note: The archive information is fully accessible via the `schema_info` command (I have tested this). Moreover, the indentation exposed an issue with some of the test utilities in `roman_datamodels`, so this PR depends on spacetelescope/roman_datamodels#96 in order for all the tests to pass everywhere.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@
 
 - Add ``IPAC/SSC`` to ``origin`` enum. [#160]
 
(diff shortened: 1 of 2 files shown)


ddavis-stsci (approved):
LGTM

PR #163, opened 2022-09-09
Rad 72 model container

PaulHuwe:
Initial schema idea for Model Containers.


PaulHuwe:
This was implemented elsewhere. Closing.

PR #166, opened 2022-09-28
Rad 75 variable units

PaulHuwe:
This PR adds units metadata for all the variables in Level 1 and Level 2 output products. This PR passes the romancal repository - not requiring changes to all files.

### DIFF
--- a/src/rad/resources/manifests/datamodels-1.0.yaml
+++ b/src/rad/resources/manifests/datamodels-1.0.yaml
@@ -75,6 +75,11 @@ tags:
   title: Photometry information
   description: |-
     Photometry information
+- tag_uri: asdf://stsci.edu/datamodels/roman/tags/wfi_science_raw_units-1.0.0
+  schema_uri: asdf://stsci.edu/datamodels/roman/schemas/wfi_science_raw_units-1.0.0
+  title: WFI Level 1 Science Data Variable Units
+  description: |-
+    WFI Level 1 Science Data Variable Units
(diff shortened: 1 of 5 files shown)


ddavis-stsci (approved):
LGTM. One small nit. I'd name the titles slightly differently
variable data --> data array
and
variable var_rnoise --> variance in the read noise array
just to be more descriptive.

perrygreenfield:
Hold off on merging. We would like to think this over a bit.

kdupriestsci:
I'm confused by this.  We want to store the units in the archive, but not the actual values?  Are units likely to change between files or will they be consistent across files?  How likely is it that someone will run a query on units (either as part of the filter or part of the result set)?  If we do want to store these fields in the archive is the ScienceCommon table the place to do it?  I.e. are these fields common to all science files?

PaulHuwe:
Correct, the units are for the values of the data arrays (some ~16Million values there). Units are LIKELY to stay the same across files of the same type, but will be different type to type (e.g. L1 vs L2). I cannot speak to how likely someone will search on units - would INS know? 

Regarding the final question, we are still in discussions with INS about implementation of this, which may affect the answer. Whenever a decision has been made, I can answer that (and, of course we will discuss it in the RAD group first).

PaulHuwe:
Converted to draft to avoid accidental merging.

PaulHuwe:
We will be going a different route - so closing this PR.

PR #168, opened 2022-10-12
RAD 99: Add schema to support Roman's non-VOUnit units.

WilliamJamieson:
Adds a schema and tag to support the non-VOUnits (`DN` and `electron`) that Roman will need to serialize in ASDF.

### DIFF
--- a/src/rad/resources/schemas/unit/unit-1.0.0.yaml
+++ b/src/rad/resources/schemas/unit/unit-1.0.0.yaml
@@ -12,7 +12,7 @@ description: >
   converter.
 
 $ref: http://stsci.edu/schemas/asdf/unit/unit-1.0.0
-enum: [DN, electron]
+# enum: [DN, electron]
 
 flowStyle: block
 ...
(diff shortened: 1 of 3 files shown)


nden (approved):
lgtm

ddavis-stsci (approved):
Looks good.

src/rad/resources/schemas/unit/unit-1.0.0.yaml
```diff
@@ -0,0 +1,18 @@
...
+title: Special Units used by Roman
+description: >
+  This represents the units used by Roman which are not part of
+  [VOUnit syntax, Version 1.0](http://www.ivoa.net/documents/VOUnits/index.html).
+
+  The specific non-VOUnit units used by Roman will be checked/enforced by the
+  converter.
+
+$ref: http://stsci.edu/schemas/asdf/unit/unit-1.0.0
+# enum: [DN, electron]
```
PaulHuwe, line 15: Are you saving this line here for when enums are fixed?

PaulHuwe (approved):
LGTM.

src/rad/resources/schemas/unit/unit-1.0.0.yaml
WilliamJamieson: Correct, it also documents the expectations here.

WilliamJamieson: `enum` support won't be possible because we need to be able to support composite units like `electron / s`. This means a lot of valid strings are possible with in this.

nden:
I am going to merge this. It's not going to be used in the B* release but no code uses it so  there's no harm merging.

PR #192, opened 2022-11-14
Array Quantities

PaulHuwe:
Ticket to support quantities for data arrays in the pipeline. 

Note: I kept the number of dimensions and types in the files for future reference, depending on future updates to quantity support in asdf.


src/rad/resources/schemas/guidewindow-1.0.0.yaml
```diff
@@ -179,21 +179,21 @@ properties:
   pedestal_frames:
     title: "Reconstituted and oriented pedestal frame GW images. Dimensions: num_frames,
             num_combined_resultants (or num_uncombined_resultants), num_reads, x, y"
-    tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
-    datatype: uint16
-    ndim: 5
+    tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
+    #datatype: uint16
+    #ndim: 5
```
WilliamJamieson, line 184:
```suggestion
    tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
      value:
        tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
        datatype: uint16
        ndim: 5
```

I think this is how you can specify the information inside the quantity tag rather than just commenting it.

WilliamJamieson: I'm not sure if my comment is exactly right. Let me know if it works. If it does not, there is some combo of things that I can spend a bit of time figuring out for you.

PaulHuwe: This is resolved with William's work on asdf-astropy and his implementation of the changes above.

src/rad/resources/schemas/guidewindow-1.0.0.yaml
```diff
@@ -188,21 +188,30 @@ properties:
   pedestal_frames:
     title: "Reconstituted and oriented pedestal frame GW images. Dimensions: num_frames,
             num_combined_resultants (or num_uncombined_resultants), num_reads, x, y"
-    tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
-    datatype: uint16
-    ndim: 5
+    tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
```
WilliamJamieson, line 191: Do we know the exact unit/units involved here? If so should we try to enforce this as part of the schema?

WilliamJamieson: Same goes for the rest of the quantities involved.

schlafly (commented):
Looks good to me.

What is special about amp33 in wfi_science_raw that leads it to be the only array that gets the additional "units:" stanza?

WilliamJamieson:
> Looks good to me.
> 
> What is special about amp33 in wfi_science_raw that leads it to be the only array that gets the additional "units:" stanza?

I see special archive keywords there, so that maybe it. Though they look new in this PR.

src/rad/resources/schemas/wfi_science_raw-1.0.0.yaml
```diff
@@ -13,14 +13,30 @@ properties:
...
+    units:
+      title: Units for the variable amp33.
+      type: string
+      sdf:
+        special_processing: VALUE_REQUIRED
+        source:
+          origin: TBD
+      archive_catalog:
+        datatype: nvarchar(10)
+        destination: [ScienceCommon.unit_level_1_amp33]
```
PaulHuwe, line 39:
```suggestion
    units:
      title: Units for the variable amp33.
      type: string
      sdf:
        special_processing: VALUE_REQUIRED
        source:
          origin: TBD
      archive_catalog:
        datatype: nvarchar(10)
        destination: [ScienceCommon.unit_level_1_amp33]
```

PaulHuwe:
> Looks good to me.
> 
> What is special about amp33 in wfi_science_raw that leads it to be the only array that gets the additional "units:" stanza?

Good catch! This was vestigial code.

PR #195, opened 2022-12-06
Rad 104: Enumerate Allowed Units for Science Data Quantities

PaulHuwe:
Add units to the schemas for science data quantities to specify allowed values.

Notes:
- ASDF library requirement update is to capture enum fix.
- Some objects have multiple units supported due to unit conversion (e.g. gain application, etc.) happening to model objects. (It is possible this may need expanding depending on how quantities are converted.)
- I made the units of crmag counts / s, but would it be better to use magnitude from astropy instead?
- I am less confident about the units of the intermediate ramp arrays (slope, yint, etc.).

### DIFF
--- a/src/rad/resources/schemas/ramp-1.0.0.yaml
+++ b/src/rad/resources/schemas/ramp-1.0.0.yaml
@@ -19,6 +19,9 @@ allOf:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: float32
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN", "electron / s"]
     pixeldq:
       title: 2-D data quality array for all planes
       tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
@@ -37,6 +40,9 @@ allOf:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: float32
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN", "electron / s"]
     amp33:
       title: Amp 33 reference pixel data
       tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
@@ -45,6 +51,9 @@ allOf:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: uint16
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN"]
     border_ref_pix_left:
       title: Original border reference pixels, on left (from viewers perspective).
       tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
@@ -52,6 +61,9 @@ allOf:
         tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
         datatype: float32
         ndim: 3
+      unit:
+        tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+        enum: ["DN"]
     border_ref_pix_right:
       title: Original border reference pixels, on right (from viewers perspective).
       tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
@@ -60,6 +72,9 @@ allOf:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: float32
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN"]
     border_ref_pix_top:
       title: Original border reference pixels, on top.
       tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
@@ -68,6 +83,9 @@ allOf:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: float32
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN"]
     border_ref_pix_bottom:
       title: Original border reference pixels, on bottom.
       tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
@@ -76,6 +94,9 @@ allOf:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: float32
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN"]
     dq_border_ref_pix_left:
       title: DQ for border reference pixels, on left (from viewers perspective).
       tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
--- a/src/rad/resources/schemas/ramp-1.0.0.yaml
+++ b/src/rad/resources/schemas/ramp-1.0.0.yaml
@@ -21,7 +21,7 @@ allOf:
           ndim: 3
         unit:
           tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
-          enum: ["DN", "electron / s"]
+          enum: ["DN", "electron"]
     pixeldq:
       title: 2-D data quality array for all planes
       tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
@@ -42,7 +42,7 @@ allOf:
           ndim: 3
         unit:
           tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
-          enum: ["DN", "electron / s"]
+          enum: ["DN", "electron"]
     amp33:
       title: Amp 33 reference pixel data
       tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
--- a/src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
+++ b/src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
@@ -13,24 +13,48 @@ allOf:
         - $ref: common-1.0.0
     slope:
       title: Segment-specific slope
-      tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
-      ndim: 3
-      datatype: float32
+      tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
+      properties:
+        value:
+          tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
+          datatype: float32
+          ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN", "electron / s"]
     sigslope:
       title: Sigma for segment-specific slope
-      tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
-      ndim: 3
-      datatype: float32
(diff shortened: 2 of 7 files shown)


ddavis-stsci:
I think in general the CR magnitude is just electrons, not counts/s, see 
https://innerspace.stsci.edu/display/JWST/ODL+Documents?preview=%2F38143003%2F40240691%2FJWST-STScI-002468.pdf

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
```diff
@@ -39,6 +63,9 @@ allOf:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: float32
           ndim: 2
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN", "electron / s"]
```
schlafly, line 68: My expectation is that yint and pedestal are the same and have the same units, so the comments above apply here as well.  I do not actually know what the distinction is between these two quantities.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
```diff
@@ -47,11 +74,16 @@ allOf:
...
+      # Possibly use u.Magnitude instead?
+      # https://docs.astropy.org/en/stable/units/logarithmic_units.html
       properties:
         value:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: float32
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["count / s"]
```
schlafly, line 86: As Dave pointed out, this is DN.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
```diff
@@ -60,6 +92,9 @@ allOf:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: float32
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["electron2 / s2"]
```
schlafly, line 97: I think these are contributions to the variance in the slope.  So if slope really can have units of DN / s (I hope not?), then these should be allowed to have units of (DN / s)^2.

src/rad/resources/schemas/ramp-1.0.0.yaml
```diff
@@ -19,6 +19,9 @@ allOf:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: float32
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN", "electron / s"]
```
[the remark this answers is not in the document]
tddesjardins: Correct, the ramp model should be in DN, while the ramp_fit_output model should have a data array in units of electrons / s.

src/rad/resources/schemas/ramp-1.0.0.yaml
```diff
@@ -45,13 +51,19 @@ allOf:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: uint16
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN"]
```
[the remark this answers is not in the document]
tddesjardins: It is correct for amp33 to be in DN. The weak lensing group requested that we keep the raw, untouched reference pixels (border + amp33) in the L2 files. So we do the reference pixel correction in DN, we remove the border, and we keep the originals in units of DN in their separate arrays.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
tddesjardins: How is crmag being used in the pipeline? I don't think it comes out in L2 data does it? So it must be something between ramp_fit_output and wfi_image.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
tddesjardins: Similar to Eddie, I'm not sure what the difference between yint and pedestal is here, but that's a different issue. Eddie is correct about the units.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
tddesjardins: If that was the case, then I agree. But since slope should only be electrons / s, then this is fine and slope should be fixed.

src/rad/resources/schemas/ramp-1.0.0.yaml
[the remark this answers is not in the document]
schlafly: Great, yeah, Stefano and I just hashed this out on #roman-dms-support.  I am skeptical that anyone will be able to use this information correctly without the L1 data, but we'll see.  Fine to keep only as DN.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
```diff
@@ -13,24 +13,48 @@ allOf:
...
-      datatype: float32
+      tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
+      properties:
+        value:
+          tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
+          datatype: float32
+          ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN", "electron / s"]
```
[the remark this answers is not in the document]
tddesjardins: Oh wait...okay, going back now that I'm thinking about it...is the order of operations here that the slope is computed in units of DN / s, and THEN the gain is applied to the slope? IF that's the case, then this is okay. The final L2 file will only be in units of electron / s, but there may be a brief window in the code in which it is in units of DN / s.

tddesjardins (commented):
See my replies to Eddie's comments. There was nothing additional other than what he and Dave pointed out. If the situation is that there is a moment in the code where the slope is DN / s before the gain is applied, but the poisson variance hasn't been computed yet, then I think that's fine. I still have questions about the use of crmag and pedestal, but those shouldn't hold up the PR.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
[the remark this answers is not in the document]
schlafly: I don't know how we actually do it, but at least in the simulator it's important to apply the gain first so that one gets the right Poisson noise.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
schlafly: I think it's auxiliary data it ramp fit output only.  It's not in the L2 data.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
```diff
@@ -13,24 +13,48 @@ allOf:
...
```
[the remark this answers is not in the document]
PaulHuwe: Sensible. Applying change.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
PaulHuwe: Applied.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
PaulHuwe: Already changed that one, thanks.

src/rad/resources/schemas/ramp-1.0.0.yaml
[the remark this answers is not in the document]
PaulHuwe: Changed.. I left it in incase the output of the step existed in RampModel at all before ImageModel, but I just double checked and it does not.

src/rad/resources/schemas/ramp-1.0.0.yaml
```diff
@@ -37,6 +40,9 @@ allOf:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: float32
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN", "electron / s"]
```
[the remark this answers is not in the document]
PaulHuwe: Changed.

PaulHuwe:
One thing to keep in mind when looking at units in these places is when the quantities get validated - which is why I left a few a bit more open (list vs just the expected output). With further regression style testing, these could be tightened.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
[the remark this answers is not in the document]
PaulHuwe: I only had it in DN in case the ramp step initialized the slope step array in STCAL with the DN input, but I just realized that it is stripped of units before passing to STCAL, so I can simplify these.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
```diff
@@ -47,11 +74,16 @@ allOf:
...
+      # Possibly use u.Magnitude instead?
+      # https://docs.astropy.org/en/stable/units/logarithmic_units.html
       properties:
         value:
           tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
           datatype: float32
           ndim: 3
+        unit:
+          tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+          enum: ["DN"]
```
schlafly, line 86: Ugh, sorry, yes, electron!  Otherwise these all look good now; thanks!

kdupriestsci (approved):
looks fine to me

WilliamJamieson (changes requested):

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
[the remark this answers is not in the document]
PaulHuwe: This is actually outdated - already changed.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
```diff
@@ -47,11 +74,16 @@ allOf:
     crmag:
       title: Approximate CR magnitudes
       tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
+      # Possibly use u.Magnitude instead?
+      # https://docs.astropy.org/en/stable/units/logarithmic_units.html
```
schlafly, line 78: We can delete this comment.

schlafly (approved):
Looks good to me.  I marked an extraneous comment I think we could remove.

src/rad/resources/schemas/ramp_fit_output-1.0.0.yaml
PaulHuwe, line 78:
```suggestion
      # Possibly use u.Magnitude instead?
      # https://docs.astropy.org/en/stable/units/logarithmic_units.html
```

PaulHuwe: Removed.

PR #196, opened 2022-12-15
move configuration into `pyproject.toml`

zacharyburnett:


pyproject.toml
```diff
@@ -31,6 +31,8 @@ docs = [
     'matplotlib',
     'docutils',
     'stsci-rtd-theme',
+    'tomli',
+    'importlib-metadata',
```
zacharyburnett, line 35: should these be conditional on Python version?

WilliamJamieson: This is fine with me

WilliamJamieson (approved):

PR #198, opened 2023-01-04
RAD-105: Update Reference Files to Quantities

PaulHuwe:
Update Reference file schemas to utilize quantities for all relevant arrays.


kdupriestsci (approved):
Looks ok to me

ddavis-stsci (approved):
LGTM

src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
```diff
@@ -38,19 +38,31 @@ properties:
       - $ref: ref_optical_element-1.0.0
   data:
     title: Dark current array
```
schlafly, line 40: Do we have a separate comment field?  I think it would be helpful to include somewhere that "dark current" is a bit of a misnomer; the dark current array actually includes the integrated number of dark counts expected in each resultant.  It's not really a dark current array, but a dark count array.  But that's not standard terminology anywhere AFAIK.

src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
```diff
@@ -38,19 +38,31 @@ properties:
...
+      unit:
+        tag: asdf://stsci.edu/datamodels/roman/tags/unit-1.0.0
+        enum: ["DN"]
   dq:
     title: 2-D data quality array for all planes
     tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
     datatype: uint32
     ndim: 2
   err:
     title: Error array
```
schlafly, line 56: I think this is the uncertainty in the dark count image; "error array" is a bit abrupt.

src/rad/resources/schemas/reference_files/distortion-1.0.0.yaml
```diff
@@ -14,12 +14,14 @@ properties:
         properties:
           reftype:
             enum: [DISTORTION]
-          output_units:
-            title: Output units of coordinates after the model is applied.
-            tag: tag:stsci.edu:asdf/unit/unit-1.0.0
           input_units:
             title: Units of the inputs to this model.
```
schlafly, line 18: How about "detector coordinates"?

src/rad/resources/schemas/reference_files/distortion-1.0.0.yaml
```diff
@@ -14,12 +14,14 @@ properties:
...
             tag: tag:stsci.edu:asdf/unit/unit-1.0.0
+            enum: ["pixel"]
+          output_units:
+            title: Output units of coordinates after the model is applied.
```
schlafly, line 22: How about V2/V3 coordinates?

src/rad/resources/schemas/reference_files/linearity-1.0.0.yaml
```diff
@@ -17,10 +17,16 @@ properties:
             enum: [LINEARITY]
   coeffs:
     title: Linearity coefficients
```
schlafly, line 19: If there's a place for a comment here, we'd want something like how this is a polynomial taking DN -> DN; it's not literally the case that all of these coefficients have units of DN.  I don't know if input_units & output_units is something that one can specify in any bit of the schema, or only on -type objects, but that would be more appropriate than making these all DN.  Sorry, I know we already discussed this, but I'm thinking ahead to stripping the DN, making the polynomials, and putting the DN back, which isn't really helping anything.

schlafly (approved):
This all looks correct to me.  I added some review comments to flesh out the "titles" of some of these arrays.  I don't know how much information we want to put there.  If there were a separate comment field I would provide some additional information beyond what we currently have.

src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
PaulHuwe: Sure! We have the `description` keyword that we can use for this (eg: https://github.com/spacetelescope/rad/blob/main/src/rad/resources/schemas/exposure-1.0.0.yaml#L36
)

PaulHuwe: I added a description.

src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
PaulHuwe: Modified.

src/rad/resources/schemas/reference_files/linearity-1.0.0.yaml
PaulHuwe: Added a description saying this.

src/rad/resources/schemas/reference_files/distortion-1.0.0.yaml
PaulHuwe: Changed.

src/rad/resources/schemas/reference_files/linearity-1.0.0.yaml
```diff
@@ -17,6 +17,9 @@ properties:
...
+    description: |
```
tddesjardins, line 20: Suggest change to: "Contains the coefficients of a polynomial to correct pixel values for classic non-linearity. Both the input and output from the polynomial are in units of DN. The coefficients have units that contain various powers of DN."

tddesjardins (approved):
Just two comments about the descriptions to clear things up a bit, but otherwise LGTM.

src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
```diff
@@ -37,20 +37,36 @@ properties:
...
     title: 2-D data quality array for all planes
     tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
     datatype: uint32
     ndim: 2
   err:
-    title: Error array
-    tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
-    datatype: float32
-    ndim: 3
+    title: Uncertainty in dark current / dark count image array
```
schlafly, line 60: Following Tyler's suggestions, let's just do:
Uncertainty in dark current array.

src/rad/resources/schemas/reference_files/linearity-1.0.0.yaml
```diff
@@ -17,6 +17,9 @@ properties:
             enum: [LINEARITY]
   coeffs:
```
schlafly, line 18: Did we also plan to add input_units and output_units to this stanza?

src/rad/resources/schemas/reference_files/linearity-1.0.0.yaml
PaulHuwe: Changed

src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
```diff
@@ -37,20 +37,36 @@ properties:
       - $ref: ref_exposure_type-1.0.0
       - $ref: ref_optical_element-1.0.0
   data:
-    title: Dark current array
-    tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
-    datatype: float32
-    ndim: 3
+    title: Dark current / dark count array
+    description: |
+       The dark current array actually includes the integrated number of
```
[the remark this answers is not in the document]
PaulHuwe: Changed.

src/rad/resources/schemas/reference_files/dark-1.0.0.yaml
PaulHuwe: Done. Also removed from dark current title.

src/rad/resources/schemas/reference_files/linearity-1.0.0.yaml
PaulHuwe: Included.

PR #202, opened 2023-01-04
Apply `codespell` to `rad`

WilliamJamieson:
Applies the [`codespell`](https://github.com/codespell-project/codespell) to `rad` in order to automatically catch (and fix) most common spelling mistakes. Note that if `codespell` catches/fixes a word which is not misspelled, one can easily ignore the word by adding it to a new line under the `ignore-words-list` in the `pyproject.toml` file.

Note this is based on #201 so that it should be merged first.


PaulHuwe (approved):
LGTM

PR #203, opened 2023-01-18
RAD-100: Write a schema for IPC reference file

PaulHuwe:
Ticket to add the interpixel capacitance correction schema.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,8 @@
 
 - Add ``pre-commit`` support. [#199]
 
+- Add IPC reference schema. [#203]
+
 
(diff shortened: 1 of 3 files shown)


alphasentaurii:
> @alphasentaurii, can you confirm that if we leave an optical element field in here (as in the current PR), we can construct the right files in CRDS such that this value is ignored until we actually have a reason to use that field?

@schlafly yes, right now the only IPC parkeys CRDS is looking for are detector and start_time, anything else would be ignored.

schlafly (approved):

schlafly:
Great.  Okay, looks good to me!

PR #204, opened 2023-01-18
RAD-78: Update observation activity description

PaulHuwe:
Adjust activity identifier in observation schema to better reflect potential values.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,7 @@
 
 - Add ``pre-commit`` support. [#199]
 
+- Adjust activity identifier in observation schema to better reflect potential values. [#204]
 
 0.14.0 (2022-11-04)
 -------------------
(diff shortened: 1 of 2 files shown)


ddavis-stsci (approved):

kdupriestsci (commented):
I don't see visit_file_activity in the database tables.  Was it renamed at some point?  Do I need to add it?
Never mind. Looking in the wrong place.

ddavis-stsci:
in Roman-STScI-000193 the activity is Hexatrigesimal so the
range is 1..ZZ (or in decimal 1..1295)

> Looks good. I think I understand the ticket to mean that it's "really" a two character string rather than an integer; I might have said 00..ZZ rather than 1..ZZ. I originally understood the change to mean that the upper limit was some not defined constant rather than the literal characters ZZ.

PaulHuwe:
> I

It isn't here?
    archive_catalog:
      datatype: nvarchar(2)
      destination: [ScienceCommon.visit_file_activity]

kdupriestsci (approved):

schlafly (approved):

PR #215, opened 2023-02-14
add source detection schema

cshanahan1:


src/rad/resources/schemas/source_detection-1.0.0.yaml
```diff
@@ -0,0 +1,13 @@
...
+$schema: asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.0.0
+id: asdf://stsci.edu/datamodels/roman/schemas/source_detection-1.0.0
+
+title: Source catalog generated by the Source Detection Step for TweakReg
+type: object
+properties:
+  tweakreg_catalog_name:
+    type: string
+  tweakreg_catalog:
+    tag: tag:astropy.org:astropy/table/table-1.0.0
```
nden, line 12: I think you need here the tag `from asdf-format/asdf-table-schemas` but I'll let @WilliamJamieson confirm the best way to use a table.

src/rad/resources/schemas/wfi_image-1.0.0.yaml
```diff
@@ -15,7 +15,9 @@ properties:
         properties:
           photometry:
             tag: asdf://stsci.edu/datamodels/roman/tags/photometry-1.0.0
-        required: [photometry]
+          source_detection:
+            tag: asdf://stsci.edu/datamodels/roman/tags/source_detection-1.0.0
```
nden, line 19: Is it necessary to note in the name that this `source_detection` section is specific to `tweak_reg`? Are we going to have a `source_detection` section for the source detection step ?

src/rad/resources/schemas/source_detection-1.0.0.yaml
WilliamJamieson: Does it have to be a table? The short answer is ASDF does not actually support writing to any of the asdf-table-schemas it only supports reading from them. They are also essentially copies of the astropy table schemas, which is what asdf uses to read a theoretically written "asdf table". So for practical reasons it should be the astropy table schema as shown here.

PaulHuwe: Is the table of a fixed format? In the past we have avoided tables in schemas and formatted them as their own objects. See https://github.com/spacetelescope/rad/blob/main/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml#L18

mairanteodoro (approved):

WilliamJamieson (approved):
These changes look good.

PR #216, opened 2023-02-21
add badges

nden:


PaulHuwe (approved):
Do you want to update the Changelog as well?

nden:
I don't think it's necessary to update the Changelog. This change does not effect the functionality.

PR #218, opened 2023-03-09
Add method to generate list of all archive_catalog destinations

WilliamJamieson:
Adds method to generate a list of all `archive_catalog` destinations.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,7 @@
 
 - Created inverse non-linearity schema. [#213]
 
-- Add ``archive.destinations`` introspection function to collect all 
+- Add ``archive.destinations`` introspection function to collect all
   ``archive_catalog.destination`` values. [#218]
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,9 @@
 
 - Created inverse non-linearity schema. [#213]
 
+- Add ``archive.destinations`` introspection function to collect all 
+  ``archive_catalog.destination`` values. [#218]
+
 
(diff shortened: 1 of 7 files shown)


WilliamJamieson:
> Would it be helpful to have the yaml keyword (possibly with path) associated with each archive_catalog entry? @ddavis-stsci @kdupriestsci

Getting anymore then the immediate keyword is impossible to get without processing the schemas through ASDF (or doing the equivalent with ad hoc logic). So this is outside the scope of what is possible with these changes.

It should be possible to get the immediate keyword; however, it will impose outside limitations on how archive_catalog values are structured (namely no internal schema logic in relation to the archive metadata) and that all  archive_catalog metadata entries are wrapped by a keyword. It will also require another layer of processing to deduce what this keyword is, which will significantly increase the code complexity.

PaulHuwe (approved):
LGTM

ddavis-stsci (approved):
Looks fine, I can run it without issues.

kdupriestsci (approved):
I'm not sure what this will be used for but it looks good to me.

WilliamJamieson:
> I'm not sure what this will be used for but it looks good to me.

This is the initial attempt to get you something that can read all the archive end-points from `rad`. So is this what you described needing last week?

kdupriestsci:
> > I'm not sure what this will be used for but it looks good to me.
> 
> This is the initial attempt to get you something that can read all the archive end-points from `rad`. So is this what you described needing last week?

This isn't quite what I'm looking for.  Please see the email I sent.

kdupriestsci (approved):

PR #219, opened 2023-03-09
import `tomli` as `tomllib` for `docs` builds before Python 3.11

zacharyburnett:
`tomli` is implemented read-only as `tomllib` in Python 3.11


WilliamJamieson:
Could you fix the code-style?

WilliamJamieson (approved):

PR #220, opened 2023-03-09
Enable use of non-VOunit tag

WilliamJamieson:
astropy/asdf-astropy#142 introduced a special tag so that non-VOunits defined within `astropy` itself could be serialized to ASDF. The `unit-1.0.0.yaml` schema introduced to `rad` in #168 was part of an attempt to be able to serialize some non-VOunits provided by `astropy`, namely `DN` and `electron`.

This PR begins the process of replacing the custom code for Roman which was introduced, partly in #168, to handle serializing and deserializing non-VOunits. @schlafly requested that we do not simply strip this support out of the Roman code base immediately because it would necessitate the regeneration of all the current reference files and invalidate any currently created Roman ASDF files. Instead, as an intermediate step, this PR simply enables using either the non-VOunit tag introduced in #168 or the new non-VOunit astropy tag from astropy/asdf-astropy#142. When the next event which requires regeneration of files occurs, the tag/schema introduced in #168 can be fully removed.


schlafly (approved):
Looks good to me.

ddavis-stsci (approved):
LGTM

PaulHuwe (approved):
LGTM. I have confirmed this passes all tests throughout the pipeline (including RCAL regression tests).

WilliamJamieson:
Once astropy/asdf-astropy#173 merges, an asdf-astropy release containing the changes needed for this can be merged.

PR #222, opened 2023-03-19
Fix schemas with $ref  at root level

eslavich:
This PR moves root-level $ref properties to an allOf combiner.  The JSON Schema spec says that when $ref is present, all other properties in that object must be ignored:

https://datatracker.ietf.org/doc/html/draft-wright-json-schema-00#section-7

Which, when the $ref is at the top of the schema, includes critical properties such as "id".  The allOf combiner will have the same validation behavior but without the potentially nasty side effects.

This will be a warning in an upcoming version of asdf, and an error at some point after that.

**Checklist**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@
 
 - Added PR Template. [#221]
 
+- Fix schemas with $ref at root level. [#222]
+
 
 0.14.1 (2023-01-31)
 -------------------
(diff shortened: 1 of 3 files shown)


WilliamJamieson (approved):
This is reasonable.

nden:
Pinging @PaulHuwe , so he's aware

PaulHuwe:
Noted, thanks.

PR #231, opened 2023-04-07
Remove use of `pytest-openfiles`

WilliamJamieson:
`pytest-openfiles` has been deprecated and is interfering with the use of fixtures in our test code (ones that handle files). This PR switches `ResourceWarnings` into errors which will catch almost all cases of files left open. The only cases it does not catch is if an open file handle is assigned to a global variable as the "warning" will not be emitted until python stops running, which occurs after pytest has completed.

Similar to spacetelescope/romancal#666

**Checklist**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+0.15.0 (unreleased)
+-------------------
+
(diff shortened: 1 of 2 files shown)


PaulHuwe (approved):
LGTM

PR #233, opened 2023-04-14
RAD-120: Read Pattern

PaulHuwe:
Resolves [RAD-120](https://jira.stsci.edu/browse/RAD-120)

Closes #

This PR adds read pattern to the exposure group.

**Checklist**
- [x] PLWishMaster Jenkins link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/127/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,8 +21,10 @@
 
 - Add database team to Code Owners file [#227]
 
-- update CodeOwners file [#230]
-  
+- Update CodeOwners file [#230]
+
+- Add read pattern to the exposure group. [#233]
+
 
 0.14.1 (2023-01-31)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+0.14.3 (unreleased)
+-------------------
+
+- Add read pattern to the exposure group. [#233]
+
+
 0.14.2 (2023-03-31)
 -------------------
 
@@ -22,8 +28,6 @@
 - Add database team to Code Owners file [#227]
 
 - Update CodeOwners file [#230]
-
-- Add read pattern to the exposure group. [#233]
 
 
 0.14.1 (2023-01-31)
--- a/src/rad/resources/schemas/exposure-1.0.0.yaml
+++ b/src/rad/resources/schemas/exposure-1.0.0.yaml
@@ -383,7 +383,7 @@ properties:
       source:
         origin: TBD
     archive_catalog:
-      datatype: nvarchar(50)
+      datatype: nvarchar(3500)
       destination: [ScienceCommon.read_pattern]
 propertyOrder: [id, type,
            start_time, mid_time, end_time,
@@ -406,6 +406,5 @@ required: [id, type,
            frame_divisor, groupgap,
            frame_time, group_time, exposure_time,
            effective_exposure_time, duration,
-           ma_table_name, ma_table_number, level0_compressed,
-           read_pattern]
+           ma_table_name, ma_table_number, level0_compressed]
 ...
--- a/src/rad/resources/schemas/exposure-1.0.0.yaml
+++ b/src/rad/resources/schemas/exposure-1.0.0.yaml
@@ -368,6 +368,23 @@ properties:
     archive_catalog:
       datatype: nchar(1)
       destination: [ScienceCommon.exposure_level0_compressed]
+  read_pattern:
+    title: Pattern of reads
+    description: |
+      Enumeration of detector reads to resultants making up the L1 data downlinked
+      from the observatory
+    type: array
+    items:
+      type: array
+      items:
+        type: integer
+    sdf:
+      special_processing: VALUE_REQUIRED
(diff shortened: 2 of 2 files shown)


src/rad/resources/schemas/exposure-1.0.0.yaml
```diff
@@ -368,6 +368,23 @@ properties:
...
+    items:
+      type: array
+      items:
+        type: integer
+    sdf:
+      special_processing: VALUE_REQUIRED
+      source:
+        origin: TBD
+    archive_catalog:
+      datatype: nvarchar(50)
```
schlafly, line 386: Can you help me with nvarchar(50)?  Does that mean a variable length string with length less than 50?  The longest possible length of this string is 3476 characters, which corresponds to a 512 read resultant where every read is assigned to its own single read resultant.

CHANGES.rst
```diff
@@ -21,8 +21,10 @@
 
 - Add database team to Code Owners file [#227]
 
-- update CodeOwners file [#230]
-  
+- Update CodeOwners file [#230]
+
+- Add read pattern to the exposure group. [#233]
+
```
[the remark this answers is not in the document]
PaulHuwe: Good catch - moved.

kdupriestsci (approved):
Looks good to me. Filed related RDADS ticket RDADS-587

ddavis-stsci:
I'm a bit confused here. 
Do we not need to preserve the grouping? For instance if the read pattern is
[1, 2], [3, 4, 5], [7, 8, 9], where 6 is skipped, turning this into an array yields,
array([1, 2, 3, 4, 5, 7, 8, 9])
but the read pattern [1, 2 ,3, 4, 5], [7, 8, 9] also gives the above array. 
So does the user find the grouping? go back to the ma table names? or does this simply not matter here?

schlafly:
I agree that the grouping is key here.  But the type is an array of arrays, which looks fine to me?

PaulHuwe:
> I agree that the grouping is key here. But the type is an array of arrays, which looks fine to me?

It should be, but I can add an explicit test for this structural preservation in RCAL-543.

ddavis-stsci:
I'm not sure that works for a list of list with an unequal number of elements, so if I try to convert the above list into an array I get 

```
np.array([[1], [2, 3], [4], [5, 6, 7, 8]])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (4,) + inhomogeneous part.
```

Am I missing something? 
I can turn that list into a 2-d array. Is that what we're proposing?
For the archive catalog that will be stored a the original list? e.g. '[1]', '[2,3]', '[4]', ...

schlafly:
Is the RAD schema going to try to push that into a numpy ndarray as you're showing?  Can we mark the type as list rather than array?  We were imagining the archive storing str([[1], [2, 3], [4], [5, 6, 7, 8]]) = '[[1], [2, 3], [4], [5, 6, 7, 8]]'

WilliamJamieson:
This cannot be an "array of arrays", that would by definition require all of the member arrays to be the same shape.

In order for this to work it must remain a "list of lists" as each constituent list can have different shape.

If we must have an array compatible format then it would be better to build a 1D array whose index represents the reads and values represent the grouping. A null flag such as -1 can be used to for skipped reads. E.G.

[1, 2], [3, 4, 5], [7, 8, 9]

Would be instead represented as

[0, 0, 1, 1, 1, -1, 2, 2, 2]

PaulHuwe:
This implementation appears to work fine?
https://github.com/spacetelescope/roman_datamodels/pull/154/files#diff-0adb667cd397bd56edce10eec11e1b10821740a10d72bd148b09645fc9108968R211

WilliamJamieson:
Note that we can do this as unsigned integers by defining 0 as the skipped positions. Then the example becomes

[1, 1, 2, 2, 2, 0, 3, 3, 3]

src/rad/resources/schemas/exposure-1.0.0.yaml
```diff
@@ -368,6 +368,23 @@ properties:
     archive_catalog:
       datatype: nchar(1)
       destination: [ScienceCommon.exposure_level0_compressed]
+  read_pattern:
+    title: Pattern of reads
+    description: |
+      Enumeration of detector reads to resultants making up the L1 data downlinked
+      from the observatory
+    type: array
```
WilliamJamieson, line 376: The confusion may also be coming from this term. Recall, the schemas are written using "jsonschema", which is based on "json" originating in JavaScript. So the "type" refers to the type you would see in JavaScript, which happens to term "array" to mean a resizable indexed collection of items which in Python we would call a "list". Hence, the schema here is describing a "list of lists" in Python.

ddavis-stsci (approved):
Looks good.

jbrookens (approved):

PR #241, opened 2023-04-21
RAD-122: L3 DataModel Schemas

PaulHuwe:
Resolves [RAD-122](https://jira.stsci.edu/browse/RAD-122)

Closes #

This PR adds level 3 mosaic datamodel schema as well as the resample metadata schema.

**Checklist**


src/rad/resources/manifests/datamodels-1.0.yaml
```diff
@@ -29,11 +29,16 @@ tags:
   title: Roman WFI Raw Science Data datamodel
   description: |-
     Basic Roman Raw Science
-- tag_uri: asdf://stsci.edu/datamodels/roman/tags/wfi_image-1.0.0
-  schema_uri: asdf://stsci.edu/datamodels/roman/schemas/wfi_image-1.0.0
+- tag_uri: asdf://stsci.edu/datamodels/roman/tags/wfi_mosaic-1.0.0
+  schema_uri: asdf://stsci.edu/datamodels/roman/schemas/wfi_mosaic-1.0.0
```
nden, line 33: I'm wondering if `mosaic` in the name reflects the correct data product since this data product is going to be used also for dithered observations, not only `mosaics`. But I don't have a better suggestion at the moment.

schlafly: And just a single FOV, yeah.  I also don't have a much better suggestion, and we can imagine that a single image is a very simple mosaic.  We could consider wfi_image_resampled or wfi_image_combined, but I like mosaic better than either of those.

tddesjardins: wfi_repixelated? 🤔

PaulHuwe: wfi_mosaic_dither

schlafly: Honestly, my vote is for keeping your original mosaic choice.  Even the single FOV case is a kind of mosaic (combining different SCAs in the same exposure).  The least mosaic-y L3 product is a long series of exposures on the exact same pointing.  If you don't like those we could retreat to something like wfi_l3, but I don't think that's more informative.

src/rad/resources/manifests/datamodels-1.0.yaml
```diff
@@ -29,11 +29,16 @@ tags:
...
   title: Wfi level 2 image information
   description: |-
```
kdupriestsci, line 35: Title and Description say level 2  but in wfi_mosaic.yaml the Title says level 3

ddavis-stsci (approved):
Besides the comments, LGTM
I'm sure this will be refined as we do more development.

src/rad/resources/manifests/datamodels-1.0.yaml
PaulHuwe: Fixed.

PaulHuwe:
If there are any objections to my merging this, please comment by COB today.

nden (approved):

PR #254, opened 2023-05-10
Fully specify the `patternProperties` values for the `wfi_img_photom` schema

WilliamJamieson:
Part of the reason why the issue leading to spacetelescope/roman_datamodels#170 was discovered was that end users were attempting to set the modified values as arrays. These values are expected to be scalar quantities with a specific unit. This PR adds additional conditions on these values to exactly specify them to be scalar quantities with a given unit.

**Checklist**

### DIFF
--- a/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
@@ -26,16 +26,34 @@ properties:
             title: Surface brightness, in MJy/steradian
             anyOf:
               - tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
+                properties:
+                  value:
+                    type: number
+                  unit:
+                    tag: tag:stsci.edu:asdf/unit/unit-1.0.0
+                    enum: [MJy.sr**-1]
               - type: "null"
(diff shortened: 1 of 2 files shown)


src/rad/resources/schemas/reference_files/wfi_img_photom-1.0.0.yaml
```diff
@@ -26,16 +26,34 @@ properties:
             title: Surface brightness, in MJy/steradian
             anyOf:
               - tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
+                properties:
+                  value:
+                    type: number
```
WilliamJamieson, line 31: This explicitly restricts the value to be a scalar, meaning it does not allow for it to be an array value. That is:
```python
import astropy.units as u

# Is valid
valid = 3.14 * (u.MJy/u.sr)

# Is invalid
invalid = [3.14] * (u.MJy/u.sr)
invalid = np.array([3.14]) * (u.MJy/u.sr)
...
```

nden (approved):

PaulHuwe (approved):
Thanks for buttoning this down. LGTM

ddavis-stsci (approved):
LGTM

PR #255, opened 2023-05-10
RAD-119 Update program to be a string

### LINKED ISSUE #251: Update program field from int to string
In the associations the program is a string not an integer and the schema should reflect that. 
An example of the program in an association is "program": "00992",

ddavis-stsci:
Resolves [RAD-119](https://jira.stsci.edu/browse/RAD-119)

Closes #251 

This PR updates the program to be a string to match association code.

**Checklist**

pytest tests/
=================================================================== test session starts ===================================================================
platform darwin -- Python 3.11.0, pytest-7.3.1, pluggy-1.0.0
rootdir: /Users/ddavis/src/Roman/rad
configfile: pyproject.toml
plugins: remotedata-0.4.0, hypothesis-6.75.1, asdf-2.15.0, mock-3.10.0, filter-subpackage-0.1.2, astropy-header-0.2.2, doctestplus-0.12.1, astropy-0.10.0, cov-4.0.0, ci-watson-0.6.1, openfiles-0.5.0, arraydiff-0.5.0
collected 397 items

tests/test_integration.py ..........................................................                                                                [ 14%]
tests/test_manifest.py .                                                                                                                            [ 14%]
tests/test_schemas.py ............................................................................................................................. [ 46%]
................................................................................................................................................... [ 83%]
..................................................................                                                                                  [100%]

=================================================================== 397 passed in 2.52s

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 0.15.0 (unreleased)
 -------------------
 
+- Update program to be a string to match association code [#255]
+
 - Update guide star id, add catalog version, and add science file name [#258]
 
 - Remove use of deprecated ``pytest-openfiles`` ``pytest`` plugin. This has been replaced by
--- a/src/rad/resources/schemas/observation-1.0.0.yaml
+++ b/src/rad/resources/schemas/observation-1.0.0.yaml
@@ -42,7 +42,7 @@ properties:
       source:
         origin: TBD
     archive_catalog:
-      datatype: smallint
+      datatype: int
       destination: [ScienceCommon.program, GuideWindow.program]
   execution_plan:
     title: Execution plan within the program, defined range is 1..99; included in obs_id and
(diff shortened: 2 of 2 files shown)


kdupriestsci (commented):
Program is widely used in the database and is expected to be an integer.  Will it cause problems if we leave it as an integer in the database and a string in the schema?

PaulHuwe (approved):
LGTM, but I defer to DB feedback on this

ddavis-stsci:
> Program is widely used in the database and is expected to be an integer. Will it cause problems if we leave it as an integer in the database and a string in the schema?

I don't think so. The association code we use has this as a string which is the motivation 
for changing this. From the ICD the program (PPPPP) has a maximum value of 18,445. So 
having this as an integer in the database should not be an issue.

kdupriestsci (approved):
Approved with program as an int for the database

PR #272, opened 2023-06-02
Bugfix for `inverse_linearity-1.0.0` `reftype`

### LINKED ISSUE #271: Invalid `reftype` in inverse_linearity
CRDS does not allow underscores in the `reftype`, this issue was not caught for inverse linearity:
https://github.com/spacetelescope/rad/blob/6f60c2cec3c68b6e0e722c5b96b72667cebc73e5/src/rad/resources/schemas/reference_files/inverse_linearity-1.0.0.yaml#L17

We will have to decide on a different and valid `reftype` for this case.

WilliamJamieson:
Resolves [RAD-126](https://jira.stsci.edu/browse/RAD-126)

Closes #271

This updates the `reftype` in the `inverse_linearity-1.0.0` schema to be compatible with CRDS. Additionally, it adds tests to the test suite to ensure future `reftype` values are CRDS compatible. 

**Checklist**
- [x] Updated relevant roman_datamodels utilities and tests (spacetelescope/roman_datamodels#195)
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-devdeps/299/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ dynamic = ['version']
 test = [
     'pytest>=4.6.0',
     'pytest-doctestplus>=0.11.1',
+    'crds>=11.16.16',
 ]
 docs = [
     'sphinx',
--- a/src/rad/resources/schemas/reference_files/inverse_linearity-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/inverse_linearity-1.0.0.yaml
@@ -14,7 +14,7 @@ properties:
         properties:
           reftype:
             type: string
-            enum: [INVERSE-LINEARITY]
+            enum: [INVERSELINEARITY]
           input_units:
             title: Units of the input to the inverse linearity polynomial.
             tag: tag:astropy.org:astropy/units/unit-1.0.0
--- a/src/rad/resources/schemas/reference_files/inverse_linearity-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/inverse_linearity-1.0.0.yaml
@@ -14,7 +14,7 @@ properties:
         properties:
           reftype:
             type: string
-            enum: [INVERSE_LINEARITY]
+            enum: [INVERSE-LINEARITY]
           input_units:
             title: Units of the input to the inverse linearity polynomial.
             tag: tag:astropy.org:astropy/units/unit-1.0.0
--- a/tests/test_schemas.py
+++ b/tests/test_schemas.py
@@ -7,10 +7,14 @@ from collections.abc import Mapping
 import asdf
 import pytest
 import yaml
+from crds.config import is_crds_name
+
+from .conftest import MANIFEST
 
 SCHEMA_URI_PREFIX = "asdf://stsci.edu/datamodels/roman/schemas/"
 METASCHEMA_URI = "asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.0.0"
 SCHEMA_URIS = [u for u in asdf.get_config().resource_manager if u.startswith(SCHEMA_URI_PREFIX) and u != METASCHEMA_URI]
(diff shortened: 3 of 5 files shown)


src/rad/resources/schemas/reference_files/inverse_linearity-1.0.0.yaml
```diff
@@ -14,7 +14,7 @@ properties:
         properties:
           reftype:
             type: string
-            enum: [INVERSE_LINEARITY]
+            enum: [INVERSE-LINEARITY]
```
WilliamJamieson, line 17: Note `INVERSELINEARITY` is also valid, please let me know which is better.

tests/test_schemas.py
```diff
@@ -150,3 +159,22 @@ def test_matched_p_exptype_entries():
...
+        if "properties" in sub_schema:
+            return sub_schema["properties"]["reftype"]["enum"][0]
+
+
+def test_reftype(ref_file_schema):
+    """
+    Check that the reftype is valid for CRDS
+    """
+    reftype = _get_reftype(ref_file_schema)
+    assert is_crds_name(f"roman_wfi_{reftype.lower()}_0000.asdf")
```
WilliamJamieson, line 180: CRDS requires a full file name. So I made up one that is valid around the inserted reftype.

tests/test_schemas.py
```diff
@@ -150,3 +159,22 @@ def test_matched_p_exptype_entries():
...
+
+def _get_reftype(schema):
+    """
+    Extract the reftype from the schema
+    """
+    all_of = schema["properties"]["meta"]["allOf"]
+
+    for sub_schema in all_of:
+        if "properties" in sub_schema:
+            return sub_schema["properties"]["reftype"]["enum"][0]
```
WilliamJamieson, line 172: The `reftype` location within the `meta` `allOf` combiner is inconsistent across all of the reference_files because they have differing numbers of entries in it.

src/rad/resources/schemas/reference_files/inverse_linearity-1.0.0.yaml
nden: My recollection is it needs to be one word, i.e. I wasn't aware `INVERSE-LINEARITY` works. Apparently it passes tests though. For consistency I would vote for `INVERSELINEARITY`

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dynamic = ['version']
 test = [
     'pytest>=4.6.0',
     'pytest-doctestplus>=0.11.1',
+    'crds>=11.16.16',
```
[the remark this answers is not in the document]
nden: We don't expect we'll need to add many new reference types (or any?) so this may be a heavy test requirement. Others should chime in whether to have this test and dependency (though really only test dependency) in RAD.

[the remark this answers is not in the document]
PaulHuwe: I do not mind the dependency. Will we be adding any ref-like files for L3? (Thinking MA Tables or other association type stuff?)

src/rad/resources/schemas/reference_files/inverse_linearity-1.0.0.yaml
PaulHuwe: I prefer INVERSELINEARITY, we haven't been separating multiple words (e.g. READNOISE).

PaulHuwe (changes requested):
LGTM with the vote for INVERSELINEARITY.

pyproject.toml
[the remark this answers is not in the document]
nden: There are no L3 reference files (at least for now). There might be par files but that's not confirmed yet.

PaulHuwe (approved):
LGTM

nden (approved):

PR #286, opened 2023-07-06
fix uri fragment in rad_schema

braingram:
This uri fragment in rad_schema is invalid:
https://github.com/spacetelescope/rad/blob/97baad66112ee72ac31ac8da52b2a19e55efd401/src/rad/resources/schemas/rad_schema-1.0.0.yaml#L102
we were getting away with this because python jsonschema < 4.18 was less strict.

See https://github.com/asdf-format/asdf-standard/pull/373 for similar changes in asdf-standard.

**Checklist**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+0.17.0 (unreleased)
+-------------------
+
+- Fix invalid uri fragment in rad_schema. [#286]
+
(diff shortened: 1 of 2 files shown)


WilliamJamieson (approved):

ddavis-stsci (approved):
LGTM

PR #296, opened 2023-07-20
Fix inverse linearity for CRDS and add uri/filename consistency checks.

### LINKED ISSUE #292: Inverse linearity schema incompatible with CRDS
CRDS certify checks asdf reference files against the schema uri (or "id"), and this check has to be valid in order for a reference file submission to be successful. The reftype/dm for "inverselinearity" is looking for an asdf tag of "inverselinearity", but the schema in RAD is still using the underscore: https://github.com/spacetelescope/rad/blob/main/src/rad/resources/manifests/datamodels-1.0.yaml. Submissions for new inverselinearity reference files are not possible until the schema is revised to match the rmaps/reference files (no underscore between inverse and linearity).

https://github.com/spacetelescope/rad/blob/main/src/rad/resources/manifests/datamodels-1.0.yaml

For reference, this is how crds.certify checks the asdf tags for Roman: https://github.com/spacetelescope/crds/blob/08da10721c0e979877dc9579b4092c79f4ceee27/crds/certify/certify.py#L496

In short, if f = the open asdf file:
f['roman']['meta']['reftype'].lower() needs to match f['roman']._tag)

### LINKED ISSUE #293: Fix Inverse Linearity RefType Discontinuity
_Issue [RAD-131](https://jira.stsci.edu/browse/RAD-131) was created on JIRA by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe):_

Issue #&#x2060;292

RefType "inverselinearity" does not match tag "inverse-linearity" and needs to be tweaked.

WilliamJamieson:
Closes #292
Closes #293

This PR fixes the inverse linearity issues persisting in CRDS. Moreover, it adds checks so that this cannot occur again for CRDS, and consistency checks for naming between the filenames, schema uris, and tag uris.

**Checklist**


WilliamJamieson:
Note that tis will create failures in the `roman_datamodels` due to the renamed `inverselinearity` schema

PaulHuwe (approved):
LGTM

PR #300, opened 2023-07-27
Add missing unit enforcements

### LINKED ISSUE #268: Missing unit/dtype declarations for some Quantities
There are several missing unit/dtype declarations which appear similar to those fixed by #254.

- The unit is specified in the title, we should just enforce it. I think this also should be a scalar value: https://github.com/spacetelescope/rad/blob/3539c2459c0e213714d272e1c0b27f2b9989ad68/src/rad/resources/schemas/photometry-1.0.0.yaml#L9C26-L13
- Same: https://github.com/spacetelescope/rad/blob/3539c2459c0e213714d272e1c0b27f2b9989ad68/src/rad/resources/schemas/photometry-1.0.0.yaml#L17-L21
- Same: https://github.com/spacetelescope/rad/blob/3539c2459c0e213714d272e1c0b27f2b9989ad68/src/rad/resources/schemas/photometry-1.0.0.yaml#L25-L29
- Same: https://github.com/spacetelescope/rad/blob/3539c2459c0e213714d272e1c0b27f2b9989ad68/src/rad/resources/schemas/photometry-1.0.0.yaml#L33-L37
- Same: https://github.com/spacetelescope/rad/blob/3539c2459c0e213714d272e1c0b27f2b9989ad68/src/rad/resources/schemas/photometry-1.0.0.yaml#L41-L45
- Same: https://github.com/spacetelescope/rad/blob/3539c2459c0e213714d272e1c0b27f2b9989ad68/src/rad/resources/schemas/photometry-1.0.0.yaml#L49-L53
- Same: https://github.com/spacetelescope/rad/blob/3539c2459c0e213714d272e1c0b27f2b9989ad68/src/rad/resources/schemas/reference_files/pixelarea-1.0.0.yaml#L21-L24
- Same: https://github.com/spacetelescope/rad/blob/3539c2459c0e213714d272e1c0b27f2b9989ad68/src/rad/resources/schemas/reference_files/pixelarea-1.0.0.yaml#L26-L29

We should at the very least enforce the units implied by the name/title for all of these. Also, if they are scalars we should explicitly make them scalars, if they are not we should specify the dtype and number of dimensions.

WilliamJamieson:
This PR adds enforcement of the units specified by descriptions of keywords, but which were not enforced by the schemas.

closes #268

**Checklist**

### DIFF
--- a/src/rad/resources/schemas/photometry-1.0.0.yaml
+++ b/src/rad/resources/schemas/photometry-1.0.0.yaml
@@ -10,6 +10,10 @@ properties:
     title: Flux density (MJy/steradian) producing 1 cps
     anyOf:
       - tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
+        properties:
+          unit:
+            tag: tag:stsci.edu:asdf/unit/unit-1.0.0
+            enum: ["MJy.sr**-1"]
       - type: "null"
     archive_catalog:
       datatype: float
@@ -18,6 +22,10 @@ properties:
     title: Flux density (uJy/arcsec2) producing 1 cps
     anyOf:
       - tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
+        properties:
+          unit:
+            tag: tag:stsci.edu:asdf/unit/unit-1.0.0
+            enum: ["uJy.arcsec**-2"]
       - type: "null"
     archive_catalog:
       datatype: float
@@ -26,6 +34,10 @@ properties:
     title: Nominal pixel area in steradians
     anyOf:
       - tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
+        properties:
+          unit:
+            tag: tag:stsci.edu:asdf/unit/unit-1.0.0
+            enum: ["sr"]
       - type: "null"
     archive_catalog:
       datatype: float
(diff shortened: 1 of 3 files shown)


src/rad/resources/schemas/photometry-1.0.0.yaml
```diff
@@ -10,6 +10,12 @@ properties:
     title: Flux density (MJy/steradian) producing 1 cps
     anyOf:
       - tag: tag:stsci.edu:asdf/unit/quantity-1.1.0
+        properties:
+          datatype:
+            enum: ["float64"]
```
WilliamJamieson, line 15: @PaulHuwe should we enforce datatype?

PaulHuwe: I like the notion, but I am concerned about it being overly restrictive (floats32 vs 64, etc.)

kdupriestsci (approved):

PaulHuwe (approved):
LGTM

src/rad/resources/schemas/photometry-1.0.0.yaml
WilliamJamieson: We can always just change this or add alternates

PR #301, opened 2023-07-27
RAD-128 Require new cal_step values

### LINKED ISSUE #287: Make newly added cal-step arguments required
_Issue [RAD-128](https://jira.stsci.edu/browse/RAD-128) was created on JIRA by [William Jamieson](https://jira.stsci.edu/secure/ViewProfile.jspa?name=wjamieson):_

These are not currently required because they would break all of the regression tests because the regression files will have to be completely rebuilt to accommodate them being required.

Thus this should wait until we are at a place we need to rebuild all the files.

WilliamJamieson:
Resolves [RAD-128](https://jira.stsci.edu/browse/RAD-128)

Closes #287

This PR makes all the newly added cal_step steps required.

**Checklist**
- [ ] Passed romancal regression testing on Jenkins / PLWishMaster. This will break all the regression tests until new files can be generated.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@
 
 - Added a prefix "s_" to the archive destination in "cal_step". [#303]
 
+- Require all the new ``cal_step`` steps to be present in the ``cal_step`` schema. [#301]
+
 0.17.0 (2023-07-27)
 -------------------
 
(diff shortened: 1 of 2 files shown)


PaulHuwe (approved):
LGTM

PR #303, opened 2023-08-02
Add archive_catalog field to ref_file

nden:

Closes #

This PR addresses an issue with archiving the reference files and populating the archive catalog. Previously the `archive_catalog` field was missing from the `ref_file` schema which caused the names of the files to not be populated in the database. In addition, there's a name collision because `reftype` is used  in `cal_step` and `ref_file`. While this is OK for the science files because they are under different tags, in the archive they go in the same database, and so for example `ref_file.dark` and `cal_step.dark` both map to `dark`. To resolve the name collision the archive destination names were changed: the ones in `cal_step` were prefixed with `s`, and the ones in `ref_file` were prefixed with `r`.

Note that these changes do not affect data files and so no changes are needed in roman_datamodels.

**Checklist**

### DIFF
--- a/src/rad/resources/schemas/cal_step-1.0.0.yaml
+++ b/src/rad/resources/schemas/cal_step-1.0.0.yaml
@@ -12,105 +12,105 @@ properties:
     enum: ['N/A', 'COMPLETE', 'SKIPPED', 'INCOMPLETE']
     archive_catalog:
       datatype: nvarchar(15)
-      destination: [ScienceRefData.assign_wcs, GuideWindow.assign_wcs]
+      destination: [ScienceRefData.s_assign_wcs, GuideWindow.s_assign_wcs]
   flat_field:
     title: Flat Field Step
     type: string
     enum: ['N/A', 'COMPLETE', 'SKIPPED', 'INCOMPLETE']
     archive_catalog:
       datatype: nvarchar(15)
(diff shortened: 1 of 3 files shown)


src/rad/resources/schemas/cal_step-1.0.0.yaml
```diff
@@ -12,105 +12,105 @@ properties:
...
-      destination: [ScienceRefData.flat_field, GuideWindow.flat_field]
+      destination: [ScienceRefData.s_flat_field, GuideWindow.s_flat_field]
   dark:
     title: Dark Subtraction
     type: string
     enum: ['N/A', 'COMPLETE', 'SKIPPED', 'INCOMPLETE']
     archive_catalog:
       datatype: nvarchar(15)
-      destination: [ScienceRefData.dark, GuideWindow.dark]
+      destination: [ScienceRefData.s_dark, GuideWindow.s_dark]
```
PaulHuwe, line 29: Do we want to send every step to GuideWindow, or just dq_init? (Since GuideWindows aren't going through all the steps, so would never have some of these?)

nden: This was left open for discussion. I think `GuideWindow` is probably not necessary at all.

kdupriestsci (approved):

PaulHuwe (approved):
LGTM

PR #334, opened 2023-10-26
RAD-138: Level 3 Mosaic Update

PaulHuwe:
Resolves [RAD-138](https://jira.stsci.edu/browse/RAD-138)

Closes #

This PR addresses ...

**Checklist**


src/rad/resources/schemas/mosaic_basic-1.0.0.yaml
```diff
@@ -0,0 +1,150 @@
...
+      special_processing: VALUE_REQUIRED
+      source:
+        origin: TBD
+    archive_catalog:
+      datatype: smallint
+      destination: [ScienceCommon.pass]
+  survey:
+    title: Observation Survey
+    type: string
+    enum: [HLS, EMS, SN, N/A]
```
schlafly, line 96: I don't think we want to limit to these strings.  In retrospect, we probably also want to add program here.  Let's discuss at 3.

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
```diff
@@ -0,0 +1,240 @@
...
+                 x0_projection, y0_projection, ra_corn1, dec_corn1, ra_corn2,
+                 dec_corn2, ra_corn3, dec_corn3, ra_corn4, dec_corn4, pixel_scale,
+                 ra_projection_center, dec_projection_center,
+                 orientat_projection_center, s_region ]
+flowStyle: block
+required: [ ra_center, dec_center, x_center, y_center, orientat, nx, ny,
+            x0_projection, y0_projection, ra_corn1, dec_corn1, ra_corn2,
+            dec_corn2, ra_corn3, dec_corn3, ra_corn4, dec_corn4, pixel_scale,
+            ra_projection_center, dec_projection_center,
+            orientat_projection_center, s_region ]
```
schlafly, line 239: Sorry, this one saw more changes than I wanted.  Can you see "Nadia's suggestion" here?
https://docs.google.com/spreadsheets/d/1sBPQRQ1pobpd9dRwfD4iGE0VJGg3LRqQCfriNTX9laQ/edit?usp=sharing
These are mostly the same fields with different names...

schlafly (commented):
This looks good.  Some things to think about for the 3 pm call:
- I don't understand very well what science.common means and we should talk about what we want there.
- The wcsinfo got a lot of name changes, sorry.
- We probably want 'program' in addition to 'survey', and let's remove the enum from survey.
- Let's talk about whether we want to temporarily give this a new name, e.g., wfi_mosaic2, to avoid breaking everything immediately.  Then migrate the L3 pipeline to use the new name, maybe?  Or some other approach.
- Let's make individual_image_meta very generic for now (no type requirements, etc, no contents), just saying that we have it.

src/rad/resources/schemas/mosaic_basic-1.0.0.yaml
PaulHuwe: Removed and program added.

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
PaulHuwe: Implemented.

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
```diff
@@ -0,0 +1,248 @@
...
+    sdf:
+      special_processing: VALUE_REQUIRED
+      source:
+        origin: TBD
+    archive_catalog:
+      datatype: float
+      destination: [ScienceCommon.dec_ref]
+  x_ref:
+    title: Pixel number in mosaic corresponding to projection center ra/dec
+    type: integer
```
schlafly, line 32: Can we make these floats / numbers?

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
```diff
@@ -0,0 +1,248 @@
...
+    sdf:
+      special_processing: VALUE_REQUIRED
+      source:
+        origin: TBD
+    archive_catalog:
+      datatype: int
+      destination: [ScienceCommon.x_ref]
+  y_ref:
+    title: Pixel number in mosaic corresponding to projection center ra/dec
+    type: integer
```
schlafly, line 42: Same here.

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
```diff
@@ -0,0 +1,248 @@
...
+    items:
+      type: string
+      enum: ["RA---TAN", "DEC--TAN"]
+    sdf:
+      special_processing: VALUE_REQUIRED
+      source:
+        origin: TBD
+    archive_catalog:
+      datatype: nvarchar(50)
+      destination: [ScienceCommon.projection]
```
schlafly, line 96: Let's just make this a string, one item.  We could make it enum: ['TAN'], though IMO we can let this be more general at the moment.

schlafly (commented):
Thanks Paul.  A few more minor changes (sorry!).  I would "approve" but I don't actually want to merge this until we look at the romancal implications.

schlafly:
Let's not make this change either yet, but the other thing I suspect from my discussion with Kim is that we'll send the metadata to a different archive destination (some L3 table rather than ScienceCommon).

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
```diff
@@ -0,0 +1,248 @@
...
+    title: "[deg] Mosaic corner 4 RA"
+    type: number
+    sdf:
+      special_processing: VALUE_REQUIRED
+      source:
+        origin: TBD
+    archive_catalog:
+      datatype: float
+      destination: [ScienceCommon.ra_corn4]
+  dec_corn4:
```
nden, line 209: Is there an advantage to having each corner ra/dec a separate attribute, instead of an array?
As disadvantage - accessing and setting them all is multiple separate calls as opposed to one assignment of an array.

Software sometimes needs these in order. What is the the order of these? It may be good to have them in the same order as WCS.footprint so it's consistent with other uses.

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
```diff
@@ -0,0 +1,248 @@
...
+    type: number
+    sdf:
+      special_processing: VALUE_REQUIRED
+      source:
+        origin: TBD
+    archive_catalog:
+      datatype: float
+      destination: [ScienceCommon.dec_corn4]
+  orientat_local:
+    title: Orientation at mosaic center
```
nden, line 220: This needs better definition - angle between what axes?

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
```diff
@@ -0,0 +1,248 @@
...
+    type: number
+    sdf:
+      special_processing: VALUE_REQUIRED
+      source:
+        origin: TBD
+    archive_catalog:
+      datatype: float
+      destination: [ScienceCommon.orientat_local]
+  orientat:
```
nden, line 229: Is this angle defining the rotation_matrix above? 
If so is it necessary, since it duplicates information and if so it needs better definition.

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
nden: I would be in favor of `TAN` as well. The rest is very FITS like. In GWCS it's not two attributes. It's either a model or a projection code. So array `["RA---TAN", "DEC--TAN"]` doesn't make sense.

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
schlafly: The argument here is to make them accessible via a SQL query in the archive.  If SQL queries support arrays better than 
I remember, an array is a good choice.

schlafly: I'm happy with matching WCS.footprint.  I don't actually know how to specify the ordering generically and would have left this undefined; I'm happy specifying something like "counterclockwise."

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
schlafly: This will be the angle between north and the y axis of the projection, turning positive in the direction of increasing right ascenscion.

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
schlafly: Yes, it's equivalent to the rotation matrix, though it's a keyword Stefano has explicitly requested.

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
PaulHuwe: Changed.

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
PaulHuwe: Changed.

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
PaulHuwe: Changed to a simple string with an enum of `TAN`

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
PaulHuwe: Updated title

src/rad/resources/schemas/mosaic_wcsinfo-1.0.0.yaml
PaulHuwe: Updated title

jbrookens (approved):

kdupriestsci (approved):
Tentatively approved, but I think we need a larger discussion on how fields for different files are stored: I don't think we should be throwing everything into ScienceCommon.

src/rad/resources/schemas/mosaic_basic-1.0.0.yaml
```diff
@@ -0,0 +1,159 @@
...
+           included in obs_id and visit_id as 'SSS'.
+    type: integer
+    sdf:
+      special_processing: VALUE_REQUIRED
+      source:
+        origin: TBD
+    archive_catalog:
+      datatype: smallint
+      destination: [ScienceCommon.segment]
+  pass:
```
WilliamJamieson, line 82: This keyword results in python code which is syntactically invalid. I suggest an alternation to the keyword to avoid this if possible.

Suggestions:

- `pass_no`
- `pass_number`

kdupriestsci:
I am concerned that wfi_mosaic-1.0.0 doesn't include basic information like filename, etc.  Why has that information been removed?

PaulHuwe:
> I am concerned that wfi_mosaic-1.0.0 doesn't include basic information like filename, etc. Why has that information been removed?

Are you referring to just the information in the basic schema?
https://github.com/spacetelescope/rad/blob/main/src/rad/resources/schemas/basic-1.0.0.yaml

In general, things not in the L3 metadata are because they are unneeded and/or inapplicable. If you have keywords that are needed for db population / MAST / etc. please recommend them.

kdupriestsci:
> > I am concerned that wfi_mosaic-1.0.0 doesn't include basic information like filename, etc. Why has that information been removed?
> 
> Are you referring to just the information in the basic schema? https://github.com/spacetelescope/rad/blob/main/src/rad/resources/schemas/basic-1.0.0.yaml
> 
> In general, things not in the L3 metadata are because they are unneeded and/or inapplicable. If you have keywords that are needed for db population / MAST / etc. please recommend them.

Yes, I was referring to the basic schema.  The database does not need these fields.  I am unsure if MAST does.  If they are not needed I'm fine with leaving them out.

ddavis-stsci (approved):
LGTM

PR #351, opened 2024-01-10
Add WCS to schemas

nden:
Part of [RCAL-749](https://jira.stsci.edu/browse/RCAL-749)

Closes #

This PR addresses ...

**Checklist**
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/559/

### DIFF
--- a/src/rad/resources/schemas/wfi_image-1.0.0.yaml
+++ b/src/rad/resources/schemas/wfi_image-1.0.0.yaml
@@ -19,7 +19,7 @@ properties:
           wcs:
             title: WCS object
             anyOf:
-              - tag: tag:stsci.edu:asdf/wcs/wcs-1.2.0
+              - tag: tag:stsci.edu:gwcs/wcs-1.1.0
               - type: "null"
           photometry:
             tag: asdf://stsci.edu/datamodels/roman/tags/photometry-1.0.0
--- a/src/rad/resources/schemas/wfi_image-1.0.0.yaml
+++ b/src/rad/resources/schemas/wfi_image-1.0.0.yaml
@@ -19,7 +19,7 @@ properties:
           wcs:
             title: WCS object
             anyOf:
-              - tag: tag:stsci.edu:gwcs/wcs-1.1.0
+              - tag: tag:stsci.edu:gwcs/wcs-1.0.0
               - type: "null"
           photometry:
             tag: asdf://stsci.edu/datamodels/roman/tags/photometry-1.0.0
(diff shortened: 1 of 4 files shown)


ddavis-stsci (approved):
LGTM

src/rad/resources/schemas/wfi_image-1.0.0.yaml
```diff
@@ -16,6 +16,11 @@ properties:
       - $ref: common-1.0.0
       - type: object
         properties:
+          wcs:
+            title: WCS object
+            anyOf:
+              - tag: tag:stsci.edu:gwcs/wcs-1.0.0
```
braingram, line 22: @nden Here's the PR that added tag 'wildcard' support:
https://github.com/asdf-format/asdf/pull/858

I am not immediately finding usage of this within rad (I was hoping to point to an existing example).

If a wildcard seems worthwhile here one option would be to restrict it to only "wcs-1.x.x" versions (in case there is a "2.0.0" that breaks things.
```suggestion
              - tag: tag:stsci.edu:gwcs/wcs-1.*
```
The more liberal option would be to use "wcs-*". I did not test either of these locally so I'm not sure if there are other issues that might be caused by the use of a wildcard here.

nden: Thanks !This works and regression tests now pass. Reading files with various WCS versions works as well.

nden:
Passing regression tests
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/550/

src/rad/resources/schemas/wfi_image-1.0.0.yaml
```diff
@@ -16,6 +16,11 @@ properties:
       - $ref: common-1.0.0
       - type: object
         properties:
+          wcs:
+            title: WCS object
+            anyOf:
+              - tag: tag:stsci.edu:gwcs/wcs-*
+              - type: "null"
```
WilliamJamieson, line 23: Why is this null? It is not a required field by the object.

WilliamJamieson (changes requested):

PaulHuwe (approved):
LGTM

src/rad/resources/schemas/wfi_image-1.0.0.yaml
nden: Thanks for pointing this out. It needs to be a required field.

WilliamJamieson: Should it still be null, or do we always expect a gwcs object in the end?

WilliamJamieson (approved):

nden:
Merging -  assuming this doesn't need to be approved for the database team

PR #374, opened 2024-02-14
Adding schema for Source Catalog

bmorris3:
### (work in progress)

This PR implements a basic schema for the output from the Source Catalog Step being developed in https://github.com/spacetelescope/romancal/pull/1102, with data models implemented in https://github.com/spacetelescope/roman_datamodels/pull/317. 

**Checklist**


src/rad/resources/schemas/source_catalog-1.0.0.yaml
```diff
@@ -0,0 +1,20 @@
+%YAML 1.1
+---
+$schema: asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.0.0
+id: asdf://stsci.edu/datamodels/roman/schemas/source_catalog-1.0.0
+
+title: Source catalog generated by the Source Catalog Step.
+datamodel_name: SourceCatalogModel
+archive_meta: None
+type: object
```
WilliamJamieson, line 9:
```suggestion
title: Source catalog generated by the Source Catalog Step.

datamodel_name: SourceCatalogModel
archive_meta: None

type: object
```

Makes it a bit easier to read.

src/rad/resources/schemas/source_catalog-1.0.0.yaml
```diff
@@ -0,0 +1,20 @@
...
+properties:
```
WilliamJamieson, line 10: General questions: Since this is a data model should it have a `meta` keyword? If so what metadata should be included.

WilliamJamieson (changes requested):
Initial comments

src/rad/resources/schemas/source_catalog-1.0.0.yaml
bmorris3: I don't know yet – I'll follow up when I get directions from RTB.

schlafly:
Re possibly including metadata, when I make fits catalogs for other surveys, what I do is include the image header as the FITS header for the catalog.  You could also imagine including the step arguments so as to record some information about how the catalog was created.

For FITS the headers are usually pretty simple, but the asdf meta area can be more extensive so it's less clear to me that this makes as much sense there.  It's also nice if the files are easy to open; I'm not sure what number of dependencies one needs to bring in to open a file that contains one of our image headers.

bmorris3:
In our last meeting, @schlafly checked if there was any downside to copying (parts of?) metadata from the L3 image used to produce the source catalog. @WilliamJamieson confirmed that ASDF would not complain if for example, GWCS was not installed, when loading a source catalog containing a copy of the metadata from its L3 image. 

So it sounds like we're likely to include some kind of metadata, though we don't know which yet.

src/rad/resources/schemas/source_catalog-1.0.0.yaml
```diff
@@ -0,0 +1,20 @@
+%YAML 1.1
+---
+$schema: asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.0.0
+id: asdf://stsci.edu/datamodels/roman/schemas/source_catalog-1.0.0
+
+title: Source catalog generated by the Source Catalog Step.
+datamodel_name: SourceCatalogModel
+archive_meta: None
```
kdupriestsci, line 8: archive_meta is used to indicate this is the entry point for storing data in the db (e.g. wfi_image-1.0.0.yaml, guidewindow-1.0.0.yaml, etc.)  Is this such a file?

bmorris3: The honest answer is: I don't know. I'm waiting on instructions from RTB on these specs.

src/rad/resources/schemas/source_catalog-1.0.0.yaml
```diff
@@ -0,0 +1,23 @@
...
+  source_catalog:
+    title: Source Catalog
+    description: |
+       Photometry and astrometry computed in the Source Catalog Step.
+    tag: tag:astropy.org:astropy/table/table-1.*
+
+
+required: [meta, source_catalog]
+flowStyle: block
+propertyOrder: [meta, source_catalog]
```
WilliamJamieson, line 22:
```suggestion
required: [source_catalog]
flowStyle: block
propertyOrder: [source_catalog]
```
This will fix the broken tests for now.

bmorris3: I wouldn't have figured this out quickly without you. Thank you 🙏🏻

src/rad/resources/schemas/segmentation_map-1.0.0.yaml
```diff
@@ -0,0 +1,32 @@
...
+  data:
+    title: Segmentation map
+    description: |
+       Segmentation map of an image model, zeros correspond to background.
+    tag: tag:stsci.edu:asdf/core/ndarray-1.*
+    ndim: 2
+    datatype: uint32
+    exact_datatype: true
+  meta:
+    anyOf:
```
nden, line 22:
```suggestion
    allOf:
```

src/rad/resources/schemas/segmentation_map-1.0.0.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - $ref: common-1.0.0
```
nden, line 23: I don't think we need all of `common` here. In the absence of direction what to do I will list what I think we need to add to the metadata.
- remove `common`

src/rad/resources/schemas/segmentation_map-1.0.0.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - type: object
+        properties:
+          filename:
+            type: string
```
nden, line 27: Under properties add
- `basic`
- `coordinates`
- `wfi_optical_element`
- `program`
- `visit` (if a visit level catalog)

remove `filename` as it is included in `basic`

src/rad/resources/schemas/source_catalog-1.0.0.yaml
```diff
@@ -0,0 +1,29 @@
...
+    description: |
+       Photometry and astrometry computed in the Source Catalog Step.
+    tag: tag:astropy.org:astropy/table/table-1.*
+  meta:
+    anyOf:
+      - $ref: common-1.0.0
+      - type: object
+        properties:
+          segmentation_map:
+            type: string
```
nden, line 24: I think this should follow what the `segmentation_map` schema includes.

nden (commented):
I wrote what I think the initial metadata for these two schemas should be.
We can change/improve the metadata when we have more information.

src/rad/resources/schemas/source_catalog-1.0.0.yaml
kdupriestsci: @nden I'm not sure what you mean by "at the level of wfi_image".  If they are their own files and not part of the wfi_image file then having the archive_meta: None is correct (we are not yet assigning a value to this keyword yet, we're just checking for its existence.

Please note that we don't have filetypes for Segmentation Maps or Source Catalogs in the archive (unless they're going under a different name), so this will require coordination with the db (either Azure or Oak, depending on who's doing it).

nden: @kdupriestsci Sorry for the bad terminology. By "at the same level as wfi_image" I meant that these will be separate files. And thanks for clarifying the correct value in this case is `None`.

@ddavis-stsci @schlafly Could you coordinate with the rest of the team the work on the filetypes in the DB?

nden:
Of course this is the schema for the files so you can't really assign catalog.meta = image.meta but need to define them. Might as well figure out which parts make sense to be included.

schlafly:
I'm not good at reading rad schemas---I think we should try to follow the image metadata structure if we do this, and I think I read this schema as saying e.g. that the engineering_quality will be accessible as meta.engineering_quality.  Meanwhile for the images it'sat meta.visit.engineering_quality.

For L2 catalogs, the most important metadata to bring over is the photometry table that does the unit conversions; let's bring that as well.  But for L3 catalogs we don't need that.  Optional?

The next most important thing is the WCS in my opinion, so let's copy that too.

Then it becomes deeply unclear to me what we need.  'instrument' for the detector and 'exposure' for the times would cover most things.  Those are both L2s that we would make optional?

Another approach would just be to do something like one metadata keyword "image_filename" which is the filename of the image that the catalog was run on, and another metadata keyword "image_metadata" that is just given as an "object" and is identically the L2 / L3 image schema copied over, trying to avoid making a new set of metadata we need to separately track and copy...

src/rad/resources/schemas/source_catalog-1.0.0.yaml
ddavis-stsci: We should bring this up in the RAD group but I think the correct filetype will be "Science WFI Level 4".

src/rad/resources/schemas/source_catalog-1.0.0.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+archive_meta: None
+type: object
+properties:
+  source_catalog:
+    title: Source Catalog
+    description: |
+       Photometry and astrometry computed in the Source Catalog Step.
+    tag: tag:astropy.org:astropy/table/table-1.*
+  meta:
```
PaulHuwe, line 18: Could you please put meta above the binary contents for both this and `segmentation_map-1.0.0.yaml`?

bmorris3: Yikes, order matters?

PaulHuwe: Doesn't matter code-wise (nor in the end file, you.. you set that order below). More about readability of schemas.

nden:
Superseded by #393

PR #379, opened 2024-02-26
Ref links

braingram:
This PR configures `sphinx-asdf` to create links to other rad schemas when a `tag` is encountered in the schema.

For example, the current `wfi_image` schema contains a `tag` to `photometry`:
https://rad.readthedocs.io/en/latest/generated/schemas/wfi_image-1.0.0.html
However this link is currently invalid (pointing to `asdf://stsci.edu/datamodels/roman/tags/photometry-1.0.0.html`). [With this PR](https://rad--379.org.readthedocs.build/en/379/generated/schemas/wfi_image-1.0.0.html) the link now points to:
https://rad.readthedocs.io/en/latest/generated/schemas/photometry-1.0.0.html

**Checklist**

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -159,3 +159,6 @@ def setup(app):
 asdf_schema_path = "../src/rad/resources"
 # This is the prefix common to all schema IDs in this repository
 asdf_schema_standard_prefix = "schemas"
+asdf_schema_reference_mappings = [
+    ("asdf://stsci.edu/datamodels/roman/tags/", "https://rad.readthedocs.io/en/latest/generated/schemas/"),
+]


nden (approved):
This looks great! Thanks for implementing it so quickly!

braingram:
Thanks for the review!

Note that the wildcard usage of external tags (like `wcs-*`) does not appear to be compatible with sphinx-asdf generating links. This means that similar mapping entries to `asdf_schema_resource_mappings` won't allow linking to `gwcs`, `asdf-standard`, etc. I'd say this is a `sphinx-asdf` bug but I didn't immediately see a way to fix it. I'll open an issue for sphinx-asdf to track it and once fixed `rad` can be updated to generate links for non-rad schemas.

PR #396, opened 2024-03-26
RAD-139: Add Members Keyword to Resample Schema

### LINKED ISSUE #340: Add Members Keyword to Resample Schema
_Issue [RAD-139](https://jira.stsci.edu/browse/RAD-139) was created on JIRA by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe):_

The members keyword needs to be added to the resample schema. It is an array of strings.

PaulHuwe:
Resolves [RAD-139](https://jira.stsci.edu/browse/RAD-139)

Closes #340 

This PR adds the members keyword to the resample schema.

**Checklist**
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://plwishmaster.stsci.edu:8081/job/RT/job/romancal/692/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,8 @@
 
 - Add attributes under the ``basic`` schema to ``WfiMosaic.meta``. [#390]
 
+- Add Members Keyword to Resample Schema. [#396]
+
 
 0.19.0 (2024-02-09)
 -------------------
--- a/src/rad/resources/schemas/resample-1.0.0.yaml
+++ b/src/rad/resources/schemas/resample-1.0.0.yaml
@@ -29,6 +29,13 @@ properties:
     description: |
       The total exposure time for the resampled product.
     type: number
+  members:
+    title: Resample Members
+    description: |
+      Filenames of resample members
+    type: array
+    items:
+      type: string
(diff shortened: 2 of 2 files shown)


kdupriestsci (approved):

schlafly:
Do we expect the members to already be included as individual_image_meta.basic.filename?  It is good to include this somewhere, and we aren't currently populating individual_image_meta.basic.filename, but once is probably enough?

ddavis-stsci (approved):
Looks good

PaulHuwe:
> Do we expect the members to already be included as individual_image_meta.basic.filename? It is good to include this somewhere, and we aren't currently populating individual_image_meta.basic.filename, but once is probably enough?

Yes, I would expect this information to be in `individual_image_meta.basic['filename']` (`basic` is a table). The reason for this ticket is that romancal does set this keyword value (e.g. https://github.com/spacetelescope/romancal/blob/main/romancal/resample/resample.py#L311), so it should be added to the schema. I have not gone through romancal to see if removing that entirely would affect other steps or not.

schlafly:
Okay, fine to duplicate this.

PR #401, opened 2024-04-01
RAD-161: Bit Mask to Resample

### LINKED ISSUE #399: Add Bit Mask to Resample
_Issue [RAD-161](https://jira.stsci.edu/browse/RAD-161) was created on JIRA by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe):_

Add a bit_mask_used (int) keyword to resample schema.

PaulHuwe:
Resolves [RAD-161](https://jira.stsci.edu/browse/RAD-161)

Closes #399 

This PR adds a bit mask field to resample.

**Checklist**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,7 +26,9 @@
 
 - Create the flux step schema. [#395]
 
+- Add bit mask field to ``resample``. [#401]
 
+  
 0.19.0 (2024-02-09)
 -------------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,8 @@
 
 - Add attributes under the ``basic`` schema to ``WfiMosaic.meta``. [#390]
 
+- Split cal_step into L2 and L3 versions. [#397]
+
 - Add Members Keyword to Resample Schema. [#396]
 
 - Create the flux step schema. [#395]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,4 @@
-0.19.1 (unreleased)
+0.20.0 (unreleased)
 -------------------
 
 - Add new schemas to documentation. [#386]
@@ -27,6 +27,8 @@
 - Create the flux step schema. [#395]
 
 - Add bit mask field to ``resample``. [#401]
+
+- Add source_catalog and segmentation_map schemas for Level 2 and Level 3 files. [#393]
 
   
 0.19.0 (2024-02-09)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,8 +30,8 @@
 
 - Add source_catalog and segmentation_map schemas for Level 2 and Level 3 files. [#393]
 
-  
-0.19.0 (2024-02-09)
+
+  0.19.0 (2024-02-09)
 -------------------
 
 - Added streamlined Level 3 Mosaic metadata schemas. [#334]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,7 +26,7 @@
 
 - Create the flux step schema. [#395]
 
-- Add bit mask field to ``resample``. [#401]
+- Create ``outlier_detection`` schema and add bit mask field to both it and ``resample``. [#401]
 
 - Add source_catalog and segmentation_map schemas for Level 2 and Level 3 files. [#393]
 
--- a/src/rad/resources/schemas/outlier_detection-1.0.0.yaml
+++ b/src/rad/resources/schemas/outlier_detection-1.0.0.yaml
@@ -1,0 +1,17 @@
+%YAML 1.1
+---
+$schema: asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.0.0
+id: asdf://stsci.edu/datamodels/roman/schemas/outlier_detection-1.0.0
+
+title: Outlier Detection Information
+type: object
+properties:
+  good_bits:
+    title: Bit Mask
+    description: |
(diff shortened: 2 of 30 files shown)


ddavis-stsci (commented):
I'm curious why is this only in resample? The first input for the good_bits string is in outlier_detection_step. Shouldn't this be in both places? Are the resample good_bits independent of those in outlier detection?

kdupriestsci (approved):

PaulHuwe:
> I'm curious why is this only in resample? The first input for the good_bits string is in outlier_detection_step. Shouldn't this be in both places? Are the resample good_bits independent of those in outlier detection?

@schlafly @mairanteodoro Can you reply?

mairanteodoro:
> I'm curious why is this only in resample? The first input for the good_bits string is in outlier_detection_step. Shouldn't this be in both places? Are the resample good_bits independent of those in outlier detection?

I think @ddavis-stsci is correct; we should have the `good_bits` parameter in both steps.

Since resample is called from outlier detection step too, maybe we should have an additional attribute for the resample parameters used with that specific step?

PaulHuwe:
> > I'm curious why is this only in resample? The first input for the good_bits string is in outlier_detection_step. Shouldn't this be in both places? Are the resample good_bits independent of those in outlier detection?
> 
> I think @ddavis-stsci is correct; the `good_bits` parameter has to be present in both steps, since resample is called from outlier detection step too. Maybe we should have an additional attribute for the resample parameters used with that specific step?

What are you proposing for the additional resample keywords?

schlafly:
Thanks Dave, Paul, Mairan.  Yes, I agree with Dave & Mairan that we need this.  It's a little frustrating because I don't think there are any situations where we would want to allow these keywords to be different, but we do have them as separate steps so we should track them separately.  The resample bit is only in the L3 products but the outlier rejection occurs on L2 images.  The outlier rejection bit should definitely be optional since we don't intend any of the products we actually archive to have that keyword.  The individual image metadata in an L3 file would include it, though, in principle; I don't know if that argues for something about how it is included in the metadata.

I think I'd argue for something like meta.outlier_detection.good_bits in the WFI image schema.  If we keep this optional, as I think we should, I think we could go ahead and merge this one and make an issue for adding something parallel for outlier detection, since that wouldn't be a schema-breaking change?

mairanteodoro:
> > Thanks Dave, Paul, Mairan. Yes, I agree with Dave & Mairan that we need this. It's a little frustrating because I don't think there are any situations where we would want to allow these keywords to be different, but we do have them as separate steps so we should track them separately. The resample bit is only in the L3 products but the outlier rejection occurs on L2 images. The outlier rejection bit should definitely be optional since we don't intend any of the products we actually archive to have that keyword. The individual image metadata in an L3 file would include it, though, in principle; I don't know if that argues for something about how it is included in the metadata.
> > I think I'd argue for something like meta.outlier_detection.good_bits in the WFI image schema. If we keep this optional, as I think we should, I think we could go ahead and merge this one and make an issue for adding something parallel for outlier detection, since that wouldn't be a schema-breaking change?
> 
> @mairanteodoro and I chatted about this yesterday, and he requested we simply import the whole resample group within outlier detection (as a "sub-folder" named resample), which would be optional. Accordingly, good_bits would come along for the ride. Do you and @ddavis-stsci agree with this approach?

@PaulHuwe I think we can go with @schlafly's suggestion instead and have `meta.outlier_detection.good_bits` instead.

ddavis-stsci:
I think Eddie's suggestion might be a better approach. Resample and 
outlier detection may have different bits set by the user for specific 
science
goals and it might be good to have both.

mairanteodoro:
> @schlafly @mairanteodoro @ddavis-stsci Another wrinkle: there is no outlier_detection keyword group in RAD. Do we want to make one, or do we want to put its good_bits elsewhere?

I'd go with making an `outlier_detection` keyword, which would make sense if we were to follow a pattern of having the step name as an attribute of `meta`.

schlafly:
Yes, I'm suggesting adding it parallel to source detection.

CHANGES.rst
mairanteodoro, line 1: Should we add an entry for `outlier_detection` too?

PaulHuwe:
`outlider_detection` with `good_bits` added.

mairanteodoro (commented):
Looks good, @PaulHuwe! Thanks.
My only suggestion is to mention the new outlier detection group in CHANGES.rst.

CHANGES.rst
PaulHuwe: Thanks for noticing - updated.

jbrookens (approved):

src/rad/resources/schemas/outlier_detection-1.0.0.yaml
```diff
@@ -0,0 +1,17 @@
...
+type: object
+properties:
+  good_bits:
+    title: Bit Mask
+    description: |
+      Bit mask parameter sent to resample.
+    type: string
+propertyOrder: [good_bits]
+flowStyle: block
+required: [good_bits]
```
schlafly, line 16: Just making sure I understand---good_bits is required if outiler_detection is specified, but outlier_detection is not required in wfi_image.  I think that's what you've written and that sounds right to me, but just checking!

PaulHuwe: Correct, that is what I intended with my implementation.

ddavis-stsci (approved):
LGTM

PR #414, opened 2024-05-01
RAD-165: Separate TVAC and FPS Schemas from Main

### LINKED ISSUE #407: Standalone TVAC and FPS Schemas
_Issue [RAD-165](https://jira.stsci.edu/browse/RAD-165) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

The current TVAC and FPS schemas reuse portions of the science data schemas. This is a problem because the science data schemas still change often and this breaks TVAC data definitions. By generating standalone schemas, not referencing existing science data tags this will be avoided.

PaulHuwe:
Resolves [RAD-165](https://jira.stsci.edu/browse/RAD-165)

Closes #407 

This PR creates a suite of separate TVAC and FPS schemas to freeze them and keep them separate from main development. NOTE: archive catalog targets need updating, which I will split out into another ticket.

**Checklist**
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://github.com/spacetelescope/roman_datamodels/pull/344/files

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 0.19.4 (unreleased)
 -------------------
 
-
+- Separated TVAC and FPS schemas into their own suite of files. [#414]
 
 0.19.3 (2024-04-25)
 -------------------
--- a/src/rad/resources/manifests/datamodels-1.0.yaml
+++ b/src/rad/resources/manifests/datamodels-1.0.yaml
@@ -303,12 +303,6 @@ tags:
   title: Telescope used to acquire the data
   description: |-
     Telescope used to acquire the data
-# SSC data models
-- tag_uri: asdf://stsci.edu/datamodels/roman/tags/msos_stack-1.0.0
-  schema_uri: asdf://stsci.edu/datamodels/roman/schemas/msos_stack-1.0.0
-  title: MSOS Stack Level 3 Schema
-  description: |-
-    Level 3 schema for SSC's MSOS stack products
 # FPS Schemas
 - tag_uri: asdf://stsci.edu/datamodels/roman/tags/fps-1.0.0
   schema_uri: asdf://stsci.edu/datamodels/roman/schemas/fps-1.0.0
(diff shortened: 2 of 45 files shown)


jbrookens (approved):
The only question I had was the archive catalog destinations being a superset, but I see you're handling those updates as a future ticket. No other concerns from me.

nden:
Looks good to me.
For my understanding - since these files are stored in a different archive, is the `archive_catalog` section even needed?

jbrookens:
> Looks good to me. For my understanding - since these files are stored in a different archive, is the `archive_catalog` section even needed?
Still needed.

kdupriestsci:
It looks like there's a lot of duplicated data between the fps and tvac folders. Can any of those files be shared between the two?

I understand that you don't want to share with science data since the science schemas are in flux, but does that problem exist at the individual test level as well?

kdupriestsci:
You've left references to ScienceRefData and GuideWindow tables in the tvac and fps files. We don't have ScienceRefData for RITA (it's stored in WFICommon.  We don't have GuideWindow for RITA (my understanding is we don't have GuideWindows for test data).

PaulHuwe:
> It looks like there's a lot of duplicated data between the fps and tvac folders. Can any of those files be shared between the two?
> 
> I understand that you don't want to share with science data since the science schemas are in flux, but does that problem exist at the individual test level as well?

It was decided to freeze TVAC and FPS individually, which means we need to keep the duplicates.

PaulHuwe:
> Will the RITA schemas always be isolated from the Roman schemas? If yes, maybe Tim @relaxing could make use of the archive_meta tag (currently set to None) to distinguish Roman from RITA?

This sounds good to me.. do you want to discuss it at a RAD meeting, or just have @relaxing implement it as it is an archive only keyword?

nden (approved):

PR #423, opened 2024-05-12
RAD-171: Add statistics field to TVAC and FPS schemas

### LINKED ISSUE #412: Add `statistics` field to TVAC and FPS schemas
After some discussion with some folks at RTB we thought that it would be useful to have an additional field called `statistics` or `image_summary_statistics` containing some summary statistics about the image that could be meaningful at the time of querying. These values would be populated at conversion.

Here are a couple of implementation options:

1. Have just one field called `image_summary_statistics` and this would be a dictionary containing some key-value pairs. This field would need to be searchable but not indexed. The description for this field would be `Summary statistics for image data cube`.
2. Alternatively, have 4 fields: `mean_counts_per_sec`, `median_counts_per_sec`, `max_counts`, `min_counts`. The descriptions would be `Mean number of counts per second`; `Median number of counts per second`, `Maximum number of counts in the data cube`, `Minimum number of counts in the data cube`, respectively. The first 2 would be `Quantity` objects with units of `DN/s` and the last two `Quantity` objects with units of `DN`. All of these should be searchable, but none of them should be indexed.

Either option would work for us, happy to discuss further.

PaulHuwe:
Resolves [RAD-171](https://jira.stsci.edu/browse/RAD-171)

Closes #412 

This PR adds a statistics schema to both FPS and TVAC.

**Checklist**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@
 
 - Separated TVAC and FPS schemas into their own suite of files. [#414]
 
+- Added statistics schemas to both FPS and TVAC. [#423]
+
 
 0.19.4 (2024-05-08)
 -------------------
--- a/src/rad/resources/manifests/datamodels-1.0.yaml
+++ b/src/rad/resources/manifests/datamodels-1.0.yaml
@@ -329,6 +329,11 @@ tags:
   title: FPS Guidestar information
   description: |-
     FPS Guidestar information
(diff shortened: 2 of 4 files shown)


fjaviersanchez (approved):
Looks good to me! Thank you

tddesjardins (approved):

jbrookens (approved):

PR #432, opened 2024-08-05
RAD-172 Sky Background

### LINKED ISSUE #247: Add schema for background
https://github.com/spacetelescope/romancal/pull/687 implements sky matching for `romancal`, and updates a (temporary) "background" dictionary in `ImageModel.meta`. The dict currently has three keys:
* `level` (`astropy.units.Quantity`)
* `subtracted` (`bool`)
* `method` (`str`)

`method` can have the options defined by the spec:
```
option('local', 'global', 'match', 'global+match', default='match')
```
This should be implemented with a schema in rad.

PaulHuwe:
Resolves [RAD-172](https://jira.stsci.edu/browse/RAD-172)

Closes #247

This PR adds a sky background schema.

**Checklist**
- [x] Passed romancal regression testing on Jenkins / PLWishMaster. Link: https://github.com/spacetelescope/RegressionTests/actions/runs/10241364423


ddavis-stsci (approved):
LGTM

kdupriestsci (approved):

src/rad/resources/schemas/sky_background-1.0.0.yaml
```diff
@@ -0,0 +1,37 @@
...
+    description: |
+      The sky background level in DN / s.
+    anyOf:
+      - tag: tag:stsci.edu:asdf/unit/quantity-1.*
+        properties:
+          datatype:
+            enum: ["float64"]
+          unit:
+            tag: tag:astropy.org:astropy/units/unit-1.*
+            enum: ["DN / s"]
```
mairanteodoro, line 20: Shouldn't we allow the same units as for the `data` array (i.e. `enum: ["DN / s", "MJy.sr**-1"]`)?

mairanteodoro (commented):
Looks good to me! Thanks, @PaulHuwe! I only have one question about the units for the background level.

src/rad/resources/schemas/sky_background-1.0.0.yaml
PaulHuwe: You only told me `DN / s`, which is what I implemented. I am okay with expanding it (if it still works with IIM). @schlafly thoughts?

schlafly: I think technically we should only ever run sky matching on calibrated data, and so the units should always be in MJy.sr**-1.  It's possible we also need DN / s in order to accommodate some unit / regression tests, in which case allowing both isn't crazy.

src/rad/resources/schemas/sky_background-1.0.0.yaml
```diff
@@ -0,0 +1,37 @@
...
+properties:
+  level:
+    title: Sky Background Level
+    description: |
+      The sky background level in DN / s.
+    anyOf:
+      - tag: tag:stsci.edu:asdf/unit/quantity-1.*
+        properties:
+          datatype:
+            enum: ["float64"]
```
schlafly, line 17: Do we have a more permissive "float" type, out of curiosity?  I feel like we should not fail validation if an f4 is provided.

src/rad/resources/schemas/sky_background-1.0.0.yaml
PaulHuwe: Added - regression test is re-running presently.

src/rad/resources/schemas/sky_background-1.0.0.yaml
schlafly: No, let's not add more code.  This is fine if there isn't a more generic type.

PaulHuwe:
Regression tests have completely passed.

mairanteodoro:
> @mairanteodoro , I do want to understand if the DN / s is needed for now or not; it's conceptually wrong but we can defer correction until later if needed.

The `DN / s` unit was just an example (not a requirement) that I provided to @PaulHuwe so that he could start his work; **it is absolutely not necessary**, since the sky match step will work fine with either unit.

The example that I gave to @PaulHuwe was from a regression test that checks for the results from the `SkyMatchStep` only. It uses the `_cal` files from [this association file](https://bytesalad.stsci.edu:443/artifactory/roman-pipeline/dev/WFI/image/L3_regtest_asn.json) from Artifactory. The data array in those `_cal` files have all `DN / s` units and that's what's passed to `SkyMatchStep`.

However, the mosaic pipeline runs `FluxStep` before `SkyMatchStep`, which fixes the unit issue (everything after `FluxStep` is in MJy/sr).

mairanteodoro (approved):
Thanks, @PaulHuwe!

jbrookens (approved):

schlafly (approved):

PR #438, opened 2024-08-29
EXAMPLE: remove quantities from wfi

braingram:
This is an example PR for one approach at removing quantities from rad schemas (to make them more general) while retaining an easy method for making `astropy.Quantity` objects from arrays with appropriate units.

With the changes in this PR and the roman_datamodels branch here:
https://github.com/braingram/roman_datamodels/tree/no_quantity
(which updates the maker utility for the `ImageModel`)
the quantities are removed from `ImageModel` while keeping the "unit" in the schema. This would allow a user to make a quantity by calling:
```python
data_quantity = Quantity(model.data, unit=str(model.schema_info("unit")["roman"]["data"]["unit"]))
```
A helper function/method could be added to make this easier, perhaps something like:
```python
data_quantity = model.as_quantity("data")
```

### DIFF
--- a/src/rad/resources/schemas/wfi_image-1.0.0.yaml
+++ b/src/rad/resources/schemas/wfi_image-1.0.0.yaml
@@ -34,18 +34,11 @@ properties:
     description: |
       Science data, excluding border reference pixels, in DNs per second
       or MJ per steradian.
-    tag: tag:stsci.edu:asdf/unit/quantity-1.*
-    properties:
-      value:
-        tag: tag:stsci.edu:asdf/core/ndarray-1.*
-        datatype: float32
-        exact_datatype: true
-        ndim: 2
-      unit:
-        oneOf:
-          - tag: tag:stsci.edu:asdf/unit/unit-1.*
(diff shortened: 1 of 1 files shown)


schlafly:
Thanks Brett.  And confirming, with ndarrays this does not occur?  I think in my imagination python is not very smart and so something like
x = x * c
ends up making a temporary array x*c in addition to x, and then replacing x with it, while something like x *= c could in principle be smarter and change x in place with no memory cost. I think you're saying that in addition to that the code is making an additional temporary array of unknown purpose; correct?

braingram:
Thanks! The `*=` should be an improvement for the reasons you described. I thought I tried this previously but it does appear to work when I run the flux unit tests with that change. I'm not sure why the temporary array would be 2x the input size but I don't see it for `*=`.

It looks like there are improvements that could be made to flux step even with quantities. I could look around for other examples if it's helpful.

schlafly:
No need to look for other examples; even if we could tweak that step to improve performance it's still indicative of substantial, surprising inefficiency with very normal looking code.

braingram:
Closing as the example served it's purpose.

PR #448, opened 2024-09-20
Varchar go vroom

braingram:

Closes #

This PR addresses ...

## Tasks

</details


schlafly:
Thanks Brett.  That's great.  I had forgotten the read_pattern and rotation_matrix cases.  I think you're right to ignore these as those lengths were picked to be very large and it's hard to see how they will cause problems.  It's also good not to touch TVAC & FPS.

braingram:
Regtests show many errors because `dummy value` is longer than many of these `nvarchar` (and now `maxLength`). These will require roman_datamodels updates to fix. One such example is `ephemeris_reference_frame`. The schema defines:
https://github.com/spacetelescope/rad/blob/525ce48b8ec9023d1034b4d82f9dc5ff628ad8b4/src/rad/resources/schemas/ephemeris-1.0.0.yaml#L43
one less then what's needed for the filled in value:
https://github.com/spacetelescope/roman_datamodels/blob/011f123c90964738dfc1f4c89da5dd4f84e533ba/src/roman_datamodels/maker_utils/_common_meta.py#L173

schlafly:
Most of the regtest failures are ~unit test failures on romancal, rather than real regression test failures.  i.e., I think it's the case that when you finish https://github.com/spacetelescope/roman_datamodels/pull/388 , the combined PR should not need regtest updates?

braingram:
> Most of the regtest failures are ~unit test failures on romancal, rather than real regression test failures. i.e., I think it's the case that when you finish [spacetelescope/roman_datamodels#388](https://github.com/spacetelescope/roman_datamodels/pull/388) , the combined PR should not need regtest updates?

Yeah. So far it's down to 1 doctest with https://github.com/spacetelescope/roman_datamodels/pull/388 but I'm on the fence about what an appropriate default value should be for `meta.filename`. Any change to it will require updating the doctest because of https://github.com/spacetelescope/romancal/blob/96af8bbc6a6dafd48d1beb251b5209d9da875ca1/docs/roman/datamodels/models.rst?plain=1#L137 and many values that might be sensible (like `test.asdf`) fail because of one source catalog [unit test](https://github.com/spacetelescope/romancal/blob/96af8bbc6a6dafd48d1beb251b5209d9da875ca1/romancal/source_catalog/tests/test_source_catalog.py#L541) (which appends a suffix to `meta.filename` making `test.asdf` `test.asdf_cat`). I think it's work a small romancal PR to fix those issues, then the roman datamodels PR to change the default values, then this PR to enable the validation.

kdupriestsci (approved):

jbrookens (approved):

PR #452, opened 2024-09-27
RAD-175: Add ePSF, ABVegaOffset, and ApCorr Schemas

PaulHuwe:
Resolves [RAD-175](https://jira.stsci.edu/browse/RAD-175)

Closes #

This PR adds schemas for PSF, AB Vega Offsets, and Aperture Corrections.

## Tasks
  - [x] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).

</details


src/rad/resources/schemas/reference_files/abvegaoffset-1.0.0.yaml
```diff
@@ -0,0 +1,137 @@
...
+      PRISM:
+        type: object
+        properties:
+          abvega_offset:
+            title: AB-Vega Magntiude Offset
+            description: Magnitude difference between the AB and Vega magnitude
+                systems. Found by calculating the AB magnitude of Vega within
+                the PRISM optical element bandpass.
+            type: number
+        required: [abvega_offset]
```
schlafly, line 132: @tddesjardins , just for my edification, we do not currently ever plan to use these in the pipeline?  These are the integrals over the full grism data, so that if that were a magnitude we ever computed, this would be the AB-Vega magnitude offset?

src/rad/resources/schemas/reference_files/apcorr-1.0.0.yaml
```diff
@@ -0,0 +1,478 @@
...
+        type: object
+        properties:
+          ap_corrections:
+            title: Aperture Corrections
+            description: The aperture correction for each enclosed energy
+                fraction. The indexing matches that of "ee_fractions".
+            tag: tag:stsci.edu:asdf/core/ndarray-1.*
+            datatype: float64
+            exact_datatype: true
+            ndim: 1
```
schlafly, line 33: @wcschultz  , can we expand the description here a little?  Is this 1/ee_fractions?  -2.5*np.log10(1/ee_fractions)  Something else?

src/rad/resources/schemas/reference_files/apcorr-1.0.0.yaml
```diff
@@ -0,0 +1,478 @@
...
+            description: Inner radius, in pixels, to use when estimating the
+                local sky background within an annulus between this
+                radius and "sky_background_rout".
+            type: number
+          sky_background_rout:
+            title: Outer Radius for the Sky Background
+            description: Outer radius, in pixels, to use when estimating the
+                local sky background within an annulus between this
+                radius and "sky_background_rin".
+            type: number
```
schlafly, line 62: @wcschultz , do the ee_fractions and aperture corrections depend on these numbers?  Presently the background we're computing comes from large scales and I don't think we're subtracting a separate local background.  It's not clear to me that I would want to change that.  Should I see this as a request to change that?

src/rad/resources/schemas/reference_files/apcorr-1.0.0.yaml
```diff
@@ -0,0 +1,478 @@
...
```
schlafly, line 472: @PaulHuwe , there's a lot of duplication among the different filter elements.  Maybe that's fine, but if, e.g., I do get clarification on the definitions of these things, that will require all of them be updated.  I don't know if there's a more efficient way to avoid the duplication, but obviously it would be nice if there were.

src/rad/resources/schemas/reference_files/epsf-1.0.0.yaml
```diff
@@ -0,0 +1,75 @@
...
+            description: Factor by which the WFI pixels have been oversampled
+                to generate the PSF.
+            type: integer
+          effective_temperature:
+            title: Effective Temperature
+            description: Effective temperature (K) of the simulated source(s)
+                used to generate the PSF model.
+            type: array
+            items:
+                type: integer
```
schlafly, line 32: @tddesjardins , somewhere we should document things like what the metallicity & gravity of the source were---or maybe this is an ideal blackbody?  I don't know where that documentation would go.

I don't know if there's a world where we would ever expect the temperature of Roman to vary enough to lead us to have different PSFs as a function of temperature, but if there were, we would probably rue calling this "effective_temperature."

src/rad/resources/schemas/reference_files/apcorr-1.0.0.yaml
braingram: I think a combination of `$ref` and `definitions` might be helpful here. Here's an example modifying the `abvegaoffset` schema:
https://github.com/braingram/rad/commit/cc8165d143c9fa55c36e8e772bdd18e4990b36b4
I haven't checked to see if this causes issues for `roman_datamodels` (or if I made any mistakes in the schema) but the general idea is supported by asdf (and jsonschema). An example currently in use is the asdf ndarray schema: https://github.com/asdf-format/asdf-standard/blob/4f88f68abbb333f3b39207d63d89b3378644f729/resources/schemas/stsci.edu/asdf/core/ndarray-1.1.0.yaml#L220

tddesjardins (approved):

src/rad/resources/schemas/reference_files/epsf-1.0.0.yaml
schlafly: Right, I understand that these are stellar effective temperatures, I just wonder whether we're setting ourselves up for confusion if there ends up being another axis "ota_temperature" and then one starts wondering what the effective temperature is the effective temperature of.  i.e., would we save confusion if we called these "stellar_effective_temperature" instead.  If you expect we'll change the name of this down the road to color or something instead then there's no need to worry about this now, though.

src/rad/resources/schemas/reference_files/abvegaoffset-1.0.0.yaml
schlafly: I agree that the AB - Vega offsets could also go in the photom files.  If we're duplicating Webb that's fine motivation for keeping it separate, but I agree that this is very parallel to the information in the photom files.

wcschultz: It is my understanding that this is what Webb did, but I would be in favor of combining the reference files instead of having 1 very basic file.

src/rad/resources/schemas/reference_files/apcorr-1.0.0.yaml
wcschultz: These are only in this reference file to mimic what Webb did. From my understanding, these values are defined by from the ee_radii at certain ee_fractions. If they are not used in romancal, I am in favor of removing them. If we want to keep them, we could improve the descriptions to be more informative.

schlafly: Okay.  @PaulHuwe , let's leave these as is, thanks Will.

src/rad/resources/schemas/reference_files/apcorr-1.0.0.yaml
schlafly: Great.  @PaulHuwe , let's replace "The indexing matches that of ee_fractions -> , corresponding to 1 / ee_fractions."  @wcschultz , I'm a bit confused about how you see this potentially changing in the future; could you elaborate when you would expect a difference between ap_corrections and 1/ee_fractions?

src/rad/resources/schemas/reference_files/abvegaoffset-1.0.0.yaml
PaulHuwe: I would also prefer this in PHOTOM, but, it appears to be separate in JWST:
https://github.com/spacetelescope/jwst/blob/e5c73fa23d157f4aa65416b4070c2b838e978494/docs/jwst/references_general/abvegaoffset_reffile.inc

src/rad/resources/schemas/reference_files/apcorr-1.0.0.yaml
PaulHuwe: Roger.

src/rad/resources/schemas/reference_files/apcorr-1.0.0.yaml
PaulHuwe: Okay, this passed my RDM PR, so I will adopt it here.

src/rad/resources/schemas/reference_files/epsf-1.0.0.yaml
PaulHuwe: I have concerns that if we enter this in as a "model only" keyword, we will have to make an entirely different schema & datamodel for release (which may have future downstream ramifications). If RTB wishes to keep the two separate, this is fine, but if we want to be able to use both interchangeably, I support renaming in a form similar to what @schlafly suggested. @tddesjardins - what say you?

src/rad/resources/schemas/reference_files/apcorr-1.0.0.yaml
wcschultz: Sorry for the confusion. I am certainly not an expert on aperture photometry and the only other definition I am aware of is the magnitude definition you described in your first comment. I was not trying to imply we need to change the definition. I was only referring to the fact that I simply copied what JWST did and that if a new definition was identified, we could change it. Currently I do not see any reason to change from 1/ee_fractions.

src/rad/resources/schemas/reference_files/apcorr-1.0.0.yaml
PaulHuwe: I decided to make this mirror the PHOTOM style, and have made it much shorter.

src/rad/resources/schemas/reference_files/apcorr-1.0.0.yaml
PaulHuwe: Change made.

src/rad/resources/schemas/reference_files/epsf-1.0.0.yaml
```diff
@@ -0,0 +1,75 @@
...
+            title: Y Position
+            description: Y-axis position of the PSF models. The order of the
+                positions corresponds to the order of the PSFs in the psf array.
+            type: array
+            items:
+                type: number
+        required: [pixel_x, pixel_y]
+  psf:
+    title: ePSF Stamps
+    description: Postage stamps of ePSF models. The 3-dimensional array is
```
PaulHuwe, line 56:
```suggestion
    description: Postage stamps of ePSF models. The 4-dimensional array is
```

PR #480, opened 2024-10-10
RAD-88: Update observation descriptions

### LINKED ISSUE #187: Update/Add observation descriptions
_Issue [RAD-88](https://jira.stsci.edu/browse/RAD-88) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

Much of this looks good but we need to update the origin in some cases. The origin for the obs_id and visit_id are correct but for items we parse from those values are still TBD. I'd suggest something like "Derived from the obs_id" or "Parsed from the obs_id". Also we could use "From the science file attributes" if we don't want to be specific.

We can also peruse <http://www-int.stsci.edu/dsd/cns/database/denali/rps_dev/> and the dms_* to try and derive the origin of some of the other parameters, e.g. observation_label, template, and such

PaulHuwe:
Resolves [RAD-88](https://jira.stsci.edu/browse/RAD-88)

Closes #187 

This PR updates observation descriptions

## Tasks

</details

### DIFF
--- a/src/rad/resources/schemas/observation-1.0.0.yaml
+++ b/src/rad/resources/schemas/observation-1.0.0.yaml
@@ -28,6 +28,22 @@ properties:
     archive_catalog:
       datatype: nvarchar(28)
       destination: [WFIExposure.observation_id, GuideWindow.observation_id]
+  visit_id:
+    title: Visit Identifier
+    description: |
+      A unique identifier for a visit. The format is "PPPPPCCAAASSSOOOVVV" where
+      "PPPPP" is the program number, "CC" is the execution plan number, "AAA" is the pass number,
+      "SSS" is the segment number, "OOO" is the observation number, and "VVV" is the
+      visit number.
+    type: string
+    sdf:
+      special_processing: VALUE_REQUIRED
+      source:
+        origin: TBD
+    maxLength: 19
+    archive_catalog:
+      datatype: nvarchar(19)
+      destination: [WFIExposure.visit_id, GuideWindow.visit_id]
   program:
     title: Program Number
     description: |
@@ -36,14 +52,13 @@ properties:
       engineering investigation to be pursued using Roman Space
       Telescope mission resources. The allowed range of program
       numbers is 00001 to 18445 inclusive.
-    type: string
-    maxLength: 6
+    type: integer
     sdf:
       special_processing: VALUE_REQUIRED
       source:
         origin: TBD
     archive_catalog:
-      datatype: nvarchar(6)
+      datatype: int
       destination: [WFIExposure.program, GuideWindow.program]
   execution_plan:
     title: Execution Plan Number
@@ -54,14 +69,13 @@ properties:
       execution plan is required whenever there is a change in the
       program. The allowed range of execution plan numbers is 01 to 99
       inclusive.
-    type: string
-    maxLength: 3
+    type: integer
     sdf:
       special_processing: VALUE_REQUIRED
       source:
         origin: TBD
     archive_catalog:
-      datatype: nvarchar(3)
+      datatype: smallint
       destination: [WFIExposure.execution_plan, GuideWindow.execution_plan]
   pass:
     title: Pass Number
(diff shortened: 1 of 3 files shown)


src/rad/resources/schemas/observation-1.0.0.yaml
```diff
@@ -25,132 +27,150 @@ properties:
...
+      numbers is 00001 to 18445 inclusive.
     type: string
+    maxLength: 6
     sdf:
       special_processing: VALUE_REQUIRED
       source:
         origin: TBD
     archive_catalog:
-      datatype: int
+      datatype: nvarchar(6)
```
kdupriestsci, line 46: I believe we consciously decided to make program in int instead of a string in the db because it makes comparisons easier.  Do we expect program to ever contain non-numeric characters?

kdupriestsci (commented):
I'm curious why you changed a number of the integer fields to strings when the expected values are integers only. This is particularly a concern for program, but I believe it carries over to some (all?) of the other fields as well.

src/rad/resources/schemas/observation-1.0.0.yaml
schlafly: @tddesjardins , do you have a preference for strings vs. ints for all of these file name components (program, plan, pass, segment, observation, visit, group, sequence, activity, exposure)?  We were consistently using a numerical type before, and could continue doing that.  I don't recall if this was something intentional that we discussed or something incidental.  I don't feel strongly here.

tddesjardins: Hm...it's not that these contain non-numeric characters, but that they may have zero-padding. I think program in particular won't ever have a leading zero because the range for programs in operations is restricted, but other elements might, and then you have different programmatic identifiers having different data types. I'm not opposed to changing it to int, but suggesting we should consider consistency.

schlafly: The suggestion is to leave this how it is and have these always be integers, which seems like enough consistency?

They are "inconsistent" in that they are zero padded in the file name and in the obsid strings, but it's not clear to me that I would really call that inconsistency.

I think this is mostly about whether we think it's more natural for people to be writing "where exposure_number = 1" or "where exposure_number = '0001'", where I'd rather write the first one than remember how many zeros everything is padded with.

tddesjardins: Sure, we can make that change. The individual program elements as integers are easy then, but I do think that things like visit ID and obs ID etc. I think will need to stay as strings since those become the building blocks of the file name.

schlafly: Yes, I agree that visit_id and obs_id should stay strings.  Note that visit_id was removed in these changes; is that intentional?

tddesjardins: Ah, I don't think that was intentional on my part. Let me get a description drafted for that and I'll post it here for @PaulHuwe to add. I think it's good to keep since another description (observation_id) references it.

tddesjardins: @PaulHuwe :
```
visit_id:
     title: Programmatic Visit Identifier
     description: A unique identifier for a visit. The format is "PPPPPCCAAASSSOOOVVV" where
                          "PPPPP" is the program number, "CC" is the execution plan number, "AAA" is the pass number,
                          "SSS" is the segment number, "OOO" is the observation number, and "VVV" is the
                           visit number.
     dtype: string
```

PaulHuwe: Okay, I am changing these back. I have a question about `visit_file_activity`: it has been a string, because the allowed values were 01 to ZZ, whereas the new comment says the range is 01 to 99 - which would allow it to be int. @tddesjardins Which is correct?

src/rad/resources/schemas/observation-1.0.0.yaml
```diff
@@ -124,29 +135,28 @@ properties:
...
       destination: [WFIExposure.visit, GuideWindow.visit]
   visit_file_group:
     title: Visit File Group
     description: |
       The visit file group identifies the sequence group
       within the visit file. The allowed range of visit file group
       numbers is 01 to 99 inclusive.
-    type: string
+    type: integer
     maxLength: 3
```
kdupriestsci, line 153: don't need maxLength if it's an integer

src/rad/resources/schemas/observation-1.0.0.yaml
tddesjardins: According to Roman-STScI-000193, SE-01, Rev B ("Roman Programmatic Data Identification" published in May 2024) in Table 3 the visit_file_activity is defined on a range of 01 to 99.

PR #481, opened 2024-10-10
RAD-92: Update visit, l1, and l2 attribute information

### LINKED ISSUE #184: Add/Update visit attribute information
_Issue [RAD-92](https://jira.stsci.edu/browse/RAD-92) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

review and update information in the visit file schema. It appears that many of the origins should be Science Data Formatting (SDF), e.g. pointing_engdb_quality

PaulHuwe:
Resolves [RAD-92](https://jira.stsci.edu/browse/RAD-92)

Closes #184 

This PR updates visit, l1, and l2 attribute information

## Tasks

</details


src/rad/resources/schemas/wfi_image-1.0.0.yaml
```diff
@@ -15,25 +15,33 @@ properties:
       - $ref: common-1.0.0
       - type: object
         properties:
+          background:
+            tag: asdf://stsci.edu/datamodels/roman/tags/sky_background-1.0.0
+          cal_logs:
+            title: Calibration Logs
```
PaulHuwe, line 21:
```suggestion
```

src/rad/resources/schemas/wfi_image-1.0.0.yaml
```diff
@@ -15,25 +15,32 @@ properties:
       - $ref: common-1.0.0
       - type: object
         properties:
+          background:
+            tag: asdf://stsci.edu/datamodels/roman/tags/sky_background-1.0.0
+          cal_logs:
+            description: |
+              Calibration logs.
```
PaulHuwe, line 22:
```suggestion
            description: |
              Calibration logs.
```

src/rad/resources/schemas/wfi_image-1.0.0.yaml
```diff
@@ -15,25 +15,30 @@ properties:
       - $ref: common-1.0.0
       - type: object
         properties:
+          background:
+            tag: asdf://stsci.edu/datamodels/roman/tags/sky_background-1.0.0
+          cal_logs:
+            tag: asdf://stsci.edu/datamodels/roman/tags/cal_logs-1.0.0
+          cal_step:
+            title: Tracker of Calibration Pipeline Steps
```
PaulHuwe, line 23:
```suggestion
```

src/rad/resources/schemas/visit-1.0.0.yaml
```diff
@@ -6,12 +6,90 @@ id: asdf://stsci.edu/datamodels/roman/schemas/visit-1.0.0
...
+          - type: "null"
+        sdf:
+          special_processing: VALUE_REQUIRED
+          source:
+            origin: TBD
+        maxLength: 100
+        archive_catalog:
+          datatype: nvarchar(100)
+          destination: [WFIExposure.dither_primary_name, SourceCatalog.engineering_quality,
+                        SegmentationMap.engineering_quality]
```
kdupriestsci, line 29: The columns should be the same across all the tables. (I assume you want dither_primary_name?)

kdupriestsci (changes requested):

src/rad/resources/schemas/wfi_image-1.0.0.yaml
```diff
@@ -135,30 +148,34 @@ properties:
...
+      Uncalibrated, four-pixel border reference pixel cube
+      from the left-edge of the detector in the science frame
+      orientation. In the L1 uncal file data array, these are pixels
+      (z, y, x) = [:, :, :4] in a zero-indexed system.
     tag: tag:stsci.edu:asdf/unit/quantity-1.*
     properties:
       value:
         tag: tag:stsci.edu:asdf/core/ndarray-1.*
-        datatype: float32
+        datatype: uint16
```
schlafly, line 161: @tddesjardins , I hadn't appreciated we were changing the type of these.  Presently the 4k by 4k pixels get converted to floats and treated ~the same up until ramp fitting, when these are separated from the science data.  That means treating them like floats for the purposes of this image.

That seems like the right behavior to me, and I read this change as saying we shouldn't do that and should instead ignore these pixels for dark subtraction, linearity correction, etc..  That would require more work in the pipeline and I wouldn't do it here.

@PaulHuwe , I'm writing this comment in the context of border_ref_pix_left, but it applies to the other border_pix as well.

It's also true that we treat the amp33 and other pixels differently.  The amp33 pixels do _not_ get the same dark subtraction, etc., that the other pixels do, since they're not actual pixels and it's not clear we could do anything meaningful here.  I think that's the right treatment as well, but it is awkward that the different reference pixels have different data types.

tddesjardins: I thought we were holding on to a version of the reference pixels that are uncalibrated, i.e., they have not been dark subtracted, linearity corrected, etc.? The IRRC is before any other calibration step, and the intent was that these arrays are copies of the untouched reference pixels (this was a request from specific science teams via the calibration working group). So I think these are supposed to be uint16 pixels in DN. You're right we carry them around a little longer and chop them off of the image after the classical linearity correction, but these are copies of the originals not the ones we chopped off.

schlafly: Ah!  Correct, indeed, thanks.  Okay, this looks fine, sorry!  I guess I register that I think carrying these on uncalibrated is somewhat silly but that's water under the bridge.

src/rad/resources/schemas/visit-1.0.0.yaml
PaulHuwe: Thanks, fixed.

PaulHuwe:
Closing as all changes already merged.

Issue #445, opened 2024-09-20
Remove Units and Quantity Object Arrays from Reference File Schema


rgcosentino:
Please remove all astropy units from reference file schema. This will eliminate quantity objects from data arrays and units included inside meta data. The list of identified reference file schema are:

- dark
- distortion
- gain
- inverse linearity
- linearity
- pixel area
- read noise
- saturation
- wfi img photom

stscijgbot-rstdms:
_This issue is tracked on JIRA as [RAD-191](https://jira.stsci.edu/browse/RAD-191)._

nden:
https://github.com/spacetelescope/rad/pull/490

PR #490, opened 2024-10-18
Remove units from the ref files

### LINKED ISSUE #489: Dark Ref File Metadata
The dark reference file schema currently has `nframes`, `groupgap`, and `ngroups`, but these aren't used by romancal or needed for CRDS matching. Can we remove these keys?

WilliamJamieson:
Resolves [RAD-192](https://jira.stsci.edu/browse/RAD-192)

Closes #489

This PR removes the units from the reference files.

## Tasks

</details

### DIFF
--- a/src/rad/resources/schemas/reference_files/distortion-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/distortion-1.0.0.yaml
@@ -13,13 +13,17 @@ properties:
     allOf:
       - $ref: asdf://stsci.edu/datamodels/roman/schemas/reference_files/ref_common-1.0.0
       - type: object
+
+        input_units: "pixel"
+        output_units: "arcsec"
+
         properties:
           reftype:
             type: string
             enum: [DISTORTION]
       - $ref: asdf://stsci.edu/datamodels/roman/schemas/reference_files/ref_optical_element-1.0.0
   coordinate_distortion_transform:
-    title: Distortion Transform Model
+    title: Distortion Transform Model with inputs in "pixel" and outputs in "arcsec"
     description: |
       The astropy.modeling.Model instance of of the distortion transform model.
     type: object
--- a/src/rad/resources/schemas/reference_files/distortion-1.0.0.yaml
+++ b/src/rad/resources/schemas/reference_files/distortion-1.0.0.yaml
@@ -17,19 +17,6 @@ properties:
           reftype:
             type: string
             enum: [DISTORTION]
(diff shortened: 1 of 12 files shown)


schlafly:
Thanks.  This looks good.  We're currently storing the units as an optional, unused bit of the schema like this:
https://github.com/spacetelescope/rad/blob/main/src/rad/resources/schemas/wfi_image-1.0.0.yaml#L105
so at least the information is in principle present in a machine readable way.  Would you mind adding those bits?

I think that doesn't actually affect RTB's ability to build the files, though, so we could advertise these branches to RTB and they could develop the needed files?

PaulHuwe:
This will also address #445 when implemented

PaulHuwe (approved):
LGTM

src/rad/resources/schemas/reference_files/distortion-1.0.0.yaml
```diff
@@ -17,19 +17,6 @@ properties:
...
-              The pixel input units of the detector coordinate model.
-            tag: tag:stsci.edu:asdf/unit/unit-1.*
-            enum: ["pixel"]
-          output_units:
-            title: Output Model Units
-            description: |
-              The V2/V3 coordinates output units after the model is applied.
-            tag: tag:stsci.edu:asdf/unit/unit-1.*
-            enum: ["arcsec"]
-        required: [output_units, input_units]
```
nden, line 32: @PaulHuwe Was there a decision to remove `input_units`` and `output_units`? I think especially the output units can be important.

PaulHuwe: You are right that these two should remain.

WilliamJamieson: Are we actually going to use this information in the pipeline or is it simply informational?

nden: It's informational.

nden: It can go in the description but it seems easier to be in a separate key. It will also show in info() this way.

WilliamJamieson: I'll add it like we added the unit descriptors for quantities

nden: Looks good to me.

PR #514, opened 2024-11-21
RAD-160: FY24 Document Review

### LINKED ISSUE #398: FY24 Document Review
_Issue [RAD-160](https://jira.stsci.edu/browse/RAD-160) was created on JIRA by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe):_

Review RAD documentation to re-align any generated discontinuities.

PaulHuwe:
Resolves [RAD-160](https://jira.stsci.edu/browse/RAD-160)

Closes #398 

This PR updates the RAD documentation to match the current schemas and fixes broken links.

## Tasks
- [x] Update relevant docstrings and / or `docs/` page.

</details


src/rad/resources/schemas/rad_schema-1.0.0.yaml
```diff
@@ -9,7 +9,7 @@ description: |
   Data Formatting and archive catalog-related properties.
 
 allOf:
-  - $ref: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
+  - $ref: http://stsci.edu/schemas/asdf/asdf-schema-1.1.0
```
braingram, line 12: This is adopting a schema from ASDF standard 1.6.0. That's not necessarily an issue but we may want to bump some versions as this was only "stable" as of asdf 4.0 (which is brand new). See this PR and comment for more details:
https://github.com/asdf-format/asdf/pull/1744#issuecomment-2479007826
Also let me know if you have any questions but the tldr; is that it might make sense to also add an ASDF standard requirement to the datamodels manifest.

PaulHuwe: Do you favour our updating this support, or simply backporting to asdf-schema-1.0.0?

braingram: My vote is to stick to `asdf-schema-1.0.0` for now. Since the rad schemas aren't versioned this can't be handled in the "standard way" (by bumping the schema versions and using the new metaschema in the new versions).

The only downside to using the old metaschema (`1.0.0`) is that we can't yet add a datatype keyword with a `float16` value to any rad schema. As there's been no discussion of using a float16 in rad I don't think this is a downside.

The (potential) downside to using the new metaschema would be that we'd have to work through what happens if a user has old versions of asdf and asdf-standard. The `asdf-schema-1.1.0` metaschema was added in `asdf-standard==1.1.0` which was only required (as a min version) for asdf 3.2.0. RAD currently has a minimum asdf version of 2.15.2 so we could use the new metaschema and bump the minimum asdf version.

To summarize I think 2 options are:
- keep the current asdf lower pin, keep using `asdf-schema-1.0.0`
- bump the asdf lower pin (I think 3.2.0 should be sufficient) and use the new `asdf-schema-1.1.0`

schlafly: I don't think this comment should drive the decision in any way.  But I will mention that Harry is interested in adding some float16s.  I've pushed back so far and I don't think we will actually do this.  But it is good to know that this is coming to asdf.

braingram: Good to know. They are supported as of asdf 4.0 (released last month) so we could start using them now (with the accompanying pin increases).

PaulHuwe: I rolled back for now. If we roll forward, we should do it cal suite wide.

docs/conf.py
```diff
@@ -155,5 +155,48 @@ def setup(app):
...
+        "https://rad.readthedocs.io/en/latest/generated/schemas/tagged_scalars/sdf_software_version-1.0.0.html",
+    ),
+    (
+        "asdf://stsci.edu/datamodels/roman/tags/telescope-1.0.0",
+        "https://rad.readthedocs.io/en/latest/generated/schemas/tagged_scalars/telescope-1.0.0.html",
+    ),
     ("asdf://stsci.edu/datamodels/roman/tags/", "https://rad.readthedocs.io/en/latest/generated/schemas/"),
+    ("asdf://stsci.edu/datamodels/roman/schemas/", "https://rad.readthedocs.io/en/latest/generated/schemas/"),
+    ("http://stsci.edu/schemas/asdf/", "https://asdf-standard.readthedocs.io/en/latest/generated/stsci.edu/asdf/"),
+    ("tag:stsci.edu:gwcs/wcs-*", "https://asdf-standard.readthedocs.io/en/latest/generated/stsci.edu/asdf/wcs/wcs-1.0.0.html"),
```
braingram, line 201:
```suggestion
    ("tag:stsci.edu:gwcs/wcs-*", "https://asdf-wcs-schemas.readthedocs.io/en/latest/generated/gwcs/wcs-1.0.0.html"),
```
I think this is the relevant wcs schema (for ASDF standard 1.5.0, ASDF standard 1.6.0 uses wcs-1.1.0).

I wonder if there is a way to create/manage these links that won't require manually updating them when schema/tag versions change. The tag wildcard is one issue here (see https://github.com/asdf-format/sphinx-asdf/issues/92).

For the other links I wonder if it's worth reorganizing the rad tags to match the schema layout (or schemas to match the tags). I think the reason the existing reference mapping isn't picking up the schemas is that the `tagged_scalars` schemas are prefixed with `.../schemas/tagged_scalars/` whereas the tags are `.../tags/...` (no tagged_scalars).

kdupriestsci (approved):

PR #525, opened 2024-12-10
move refs under allOfs where required

braingram:
Including a `$ref` key in a schema mapping/object/dict comes with the requirement that other members of that mapping are ignored. See https://datatracker.ietf.org/doc/html/draft-pbryan-zyp-json-ref-03#section-3

This PR updates the RAD schemas nesting `$ref`s in `allOf` combiners when the mapping contains other items and adds a test to verify the changes.

Romancal regtests all pass: https://github.com/spacetelescope/RegressionTests/actions/runs/12260132209

## Tasks

</details

### DIFF
--- a/tests/test_schemas.py
+++ b/tests/test_schemas.py
@@ -297,16 +297,7 @@ def test_varchar_length(uri):
     asdf.treeutil.walk(schema, callback)
 
 
-# don't test tvac or fps schemas as they are static
-@pytest.mark.parametrize(
-    "uri",
-    [
-        uri
-        for uri in SCHEMA_URIS
-        if not uri.startswith("asdf://stsci.edu/datamodels/roman/schemas/fps/")
-        and not uri.startswith("asdf://stsci.edu/datamodels/roman/schemas/tvac/")
-    ],
-)
+@pytest.mark.parametrize("uri", SCHEMA_URIS)
 def test_ref_loneliness(uri):
     """
     An object with a $ref should contain no other items
--- a/tests/test_schemas.py
+++ b/tests/test_schemas.py
@@ -295,3 +295,29 @@ def test_varchar_length(uri):
         assert node["maxLength"] == v, f"archive_catalog nvarchar does not match maxLength in schema {uri}"
 
     asdf.treeutil.walk(schema, callback)
+
+
+# don't test tvac or fps schemas as they are static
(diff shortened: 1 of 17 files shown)


tests/test_schemas.py
```diff
@@ -295,3 +295,29 @@ def callback(node, nvarchars=None):
         assert node["maxLength"] == v, f"archive_catalog nvarchar does not match maxLength in schema {uri}"
 
     asdf.treeutil.walk(schema, callback)
+
+
+# don't test tvac or fps schemas as they are static
```
schlafly, line 300: This seems like good policy to me, but checking: this change kind of changes the syntax of the schema without changing the meaning, so we probably wouldn't break tvac / fps by updating them as well?

braingram: Yeah that's a good way of putting it.

I'm happy to add the tvac/fps schemas to this test (and update them where necessary). Let me know if you'd like me to update those.

schlafly: No, I'm happy to defer to you about which route is easier or more maintainable---I just got worried about whether this was a change that would lead to tvac/fps breakage when I saw this, which I didn't expect given the passing regtests.

braingram: Thanks! Is there any situation where a file using a tvac/fps schema (or more specifically a tvac/fps tag) will be archived?

If so, updating the schemas with similar changes could avoid a future bug when asdf fixes schema_info to handle `$ref`s as the spec requires.

If not, I'd say we keep them as is.

schlafly: I think roughly speaking archiving is the _only_ thing we do to tvac / fps data, and that a substantial amount of tvac / fps data has already been archived in RITA (Roman Integration and Test Archive).

braingram: [0fd380a](https://github.com/spacetelescope/rad/pull/525/commits/0fd380ae06c70eab9aa92344c125e50fd6824bae) adds the fps/tvac schemas to the test and updates the required `$ref`s.

schlafly (approved):
Okay, thanks Brett.  I'm not a maintainer but I'm happy to approve a bug fix that passes regtests.  This does also highlight that we don't have any tvac / fps tests in the current regtests; perhaps @stscieisenhamer can add one of those as part of his TVAC translation tests.

PR #528, opened 2025-01-06
RCAL-977 - Version datamodels

braingram:
Resolves [RCAL-977](https://jira.stsci.edu/browse/RCAL-977)

This PR prepares RAD for versioning by:
- updating docs to describe the versioning process
- updating tests to allow for multiple manifests

See https://github.com/spacetelescope/roman_datamodels/pull/445 for more details.

## Tasks

</details


docs/manifests.rst
```diff
@@ -9,3 +9,4 @@ The ``tag`` manifests:
     :standard_prefix: manifests
 
     datamodels-1.0
```
WilliamJamieson, line 11: For consistency purposes could we update `datamodels-1.0` to `datamodels-1.0.0` (file name etc)?

tests/test_manifest.py
```diff
@@ -18,21 +15,21 @@ def test_manifest_valid(manifest):
...
+
+        # Check the URIs
+        assert entry["tag_uri"].startswith("asdf://stsci.edu/datamodels/roman/tags/")
+        uri_suffix = entry["tag_uri"].split("asdf://stsci.edu/datamodels/roman/tags/")[-1]
+        # Remove tagged scalars from the uri string
+        schema_uri = entry["schema_uri"]
+        if "tagged_scalars" in schema_uri.split("/"):
+            schema_uri = schema_uri.replace("tagged_scalars/", "")
+        assert schema_uri.endswith(uri_suffix)
+        assert schema_uri.startswith("asdf://stsci.edu/datamodels/roman/schemas/")
```
WilliamJamieson, line 35: In keeping with the original parameterization, you can add the following fixture to `conftest.py`:

```python
MANIFEST_ENTRIES = []
for manifest in MANIFESTS:
    MANIFEST_ENTRIES.extend(manifest["tags"])


@pytest.fixture(scope="session", params=MANIFEST_ENTRIES)
def manifest_entry(request):
    return request.param
```

Then simply use `manifest_entry` rather than `entry` as the fixture argument to the test. If nothing else this will make it very simple to identify all the entries with issues individually instead of constantly re-running the test.

tests/test_schemas.py
```diff
@@ -11,18 +11,36 @@
...
+    uris.update(
+        [
+            "tag:stsci.edu:asdf/time/time-1.*",
+            "tag:stsci.edu:asdf/core/ndarray-1.*",
+            "tag:stsci.edu:asdf/unit/quantity-1.*",
+            "tag:stsci.edu:asdf/unit/unit-1.*",
+            "tag:astropy.org:astropy/units/unit-1.*",
+            "tag:astropy.org:astropy/table/table-1.*",
+            "tag:stsci.edu:gwcs/wcs-*",
+        ]
```
WilliamJamieson, line 39: Could we move this to the top constants with a name like `EXTERNAL_TAG_URIS`?

WilliamJamieson (changes requested):

docs/manifests.rst
braingram: Thanks! I removed `2.0.0` in c38062fe97b38cb84784f7597ec33d1afa6c1471 (since it's not needed for this PR).

tests/test_manifest.py
braingram: Thanks for the suggestion. I added the fixture and updated the test in 903463eec98b66859bfe2036e0280629649683bb

tests/test_schemas.py
```diff
@@ -11,18 +11,36 @@
...
 
-from .conftest import MANIFEST
+from .conftest import MANIFESTS
 
 SCHEMA_URI_PREFIX = "asdf://stsci.edu/datamodels/roman/schemas/"
 METASCHEMA_URI = "asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.0.0"
 SCHEMA_URIS = [u for u in asdf.get_config().resource_manager if u.startswith(SCHEMA_URI_PREFIX) and u != METASCHEMA_URI]
-REF_FILE_SCHEMA_URIS = [u["schema_uri"] for u in MANIFEST["tags"] if "/reference_files/" in u["schema_uri"]]
+TAG_DEFS = [tag_def for manifest in MANIFESTS for tag_def in manifest["tags"]]
+REF_FILE_TAG_DEFS = [tag_def for tag_def in TAG_DEFS if "/reference_files" in tag_def["schema_uri"]]
```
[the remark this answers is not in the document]
braingram: I switched these to tuples in 349a57b1463d11011f7d53bfe53f714b70c782c0

tests/test_schemas.py
braingram: Moved to the top in https://github.com/spacetelescope/rad/commit/349a57b1463d11011f7d53bfe53f714b70c782c0

WilliamJamieson (approved):

schlafly:
Brett and I talked a fair amount about wanting to merge this soon.  I think we could put this in now.

One possibility we discussed was changing the default version to 0.0.0 instead of the somewhat misleading 1.0.0.  I think that would break all of our files (?), but would otherwise be a sensible change.  Other thoughts there?  I guess we're not required to make that decision now, though we should make it soon.  Such a change could also affect the reference files, though those have been more stable than the science files.

@nden , anything else we should think about before merging this?  @braingram , anything else I'm missing that we talked about on ~Thursday?

braingram:
> Brett and I talked a fair amount about wanting to merge this soon. I think we could put this in now.
> 
> One possibility we discussed was changing the default version to 0.0.0 instead of the somewhat misleading 1.0.0. I think that would break all of our files (?), but would otherwise be a sensible change. Other thoughts there? I guess we're not required to make that decision now, though we should make it soon. Such a change could also affect the reference files, though those have been more stable than the science files.
> 
> @nden , anything else we should think about before merging this? @braingram , anything else I'm missing that we talked about on ~Thursday?

Thanks! Yeah changing the current version to 0.0.1 isn't addressed in this PR and could use some more discussion/consideration (more on that below). I did update the docs to include some developer focused descriptions of:
- schema/manifest versioning: https://rad--528.org.readthedocs.build/en/528/versioning.html#schema-versioning
- how RAD handles old versions (with this PR): https://rad--528.org.readthedocs.build/en/528/versioning.html#old-version-support
- dropping older version (but retaining the ability to open the file in asdf): https://rad--528.org.readthedocs.build/en/528/versioning.html#dropping-support-for-pre-flight-versions

For switching the current version to 0.0.1 we might want to first consider which schemas we deem to be "stable". For example if the TVAC, FPS and reference files are stable we could leave those as 1.0.0 (which would allow those files to open). The other schemas (and associated tags) could be rolled back to 0.0.1 (breaking all files that use those tags). The manifest could be rolled back to 0.0.1 (which would cause an `AsdfPackageVersionWarning` warning when opening any file written with the 1.0.0 manifest, regenerating the reference files would silence the warning). From there we could march the versions forward until things are stable enough that we can claim 1.0.0 again. A slight modification to this approach would be to reversion everything to 0.0.1 knowing that we need to regenerate reference files and TVAC/FPS files.

PR #524, opened 2024-12-06
RCAL-919: Formalize the patches (sky cells) file (e.g., add it as a reference file)

perrygreenfield:
Resolves [RCAL](https://jira.stsci.edu/browse/RCAL-919)

This PR addresses the need to support the skycell definition file as a reference file

The schema to support this file has been added to RAD.

PR #536, opened 2025-01-31
RCAL-919: Formalize the patches (sky cells) file (e.g., add it as a reference file)

perrygreenfield:
Resolves [RCAL-919](https://jira.stsci.edu/browse/RCAL-919)
Closes #1420

This PR addresses the need to support the skycell definition file as a reference file. 

The schema to support this file has been added to RAD.

This supersedes PR #524 due to my screwups in managing that. I believe it addressed most of the comments made for that.


tests/test_schemas.py
```diff
@@ -306,6 +306,9 @@ def test_ref_file_meta_common(ref_file_schema):
     """
     all_of = ref_file_schema["properties"]["meta"]["allOf"]
 
+    if ref_file_schema["id"].find("skycells") >= 0:
```
PaulHuwe, line 309: Is this expected to return a negative value at some point? Otherwise it implies the rest of the code would never be run.

src/rad/resources/schemas/reference_files/skycells-1.0.0.yaml
```diff
@@ -0,0 +1,199 @@
+%YAML 1.1
+---
+$schema: asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.0.0
+id: asdf://stsci.edu/datamodels/roman/schemas/reference_files/skycells-1.0.0
+
+title: Skycells File Schema
```
PaulHuwe, line 6:
```suggestion
title: Skycells Reference File Schema
```

tests/test_schemas.py
perrygreenfield: Yes, if the reference file schema is not a skycell schema the value should be -1 and then it should continue on; if it is, it doesn't use the common meta since some of the attributes in that make no sense for this reference file.

PaulHuwe (changes requested):

src/rad/resources/schemas/reference_files/skycells-1.0.0.yaml
perrygreenfield: Done

src/rad/resources/manifests/datamodels-1.0.yaml
```diff
@@ -513,4 +513,9 @@ tags:
   title: MSOS Stack Level 3 Schema
   description: |-
     Level 3 schema for SSC's MSOS stack products
+- tag_uri: asdf://stsci.edu/datamodels/roman/tags/reference_files/skycells-1.0.0
+  schema_uri: asdf://stsci.edu/datamodels/roman/schemas/reference_files/skycells-1.0.0
+  title: Schema for the skycell definitions
```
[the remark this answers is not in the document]
perrygreenfield: Done

src/rad/resources/manifests/datamodels-1.0.yaml
```diff
@@ -513,4 +513,9 @@ tags:
   title: MSOS Stack Level 3 Schema
   description: |-
     Level 3 schema for SSC's MSOS stack products
+- tag_uri: asdf://stsci.edu/datamodels/roman/tags/reference_files/skycells-1.0.0
```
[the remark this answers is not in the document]
nden: @perrygreenfield I think this is still under SSC models. Should go just after https://github.com/spacetelescope/rad/blob/0d949a0bf25ed5716cdf07020e93c9b4a7b0dc84/src/rad/resources/manifests/datamodels-1.0.yaml#L248

src/rad/resources/schemas/reference_files/skycells-1.0.0.yaml
```diff
@@ -0,0 +1,199 @@
...
+
+datamodel_name: SkycellsRefModel
+
+type: object
+properties:
+  meta:
+    title: Metadata for Skycells definitions reference file
+    description: |
+      The necessary metadata for the Skycells reference file
+    allOf:
```
nden, line 16: May be add here 
`- $ref: asdf://stsci.edu/datamodels/roman/schemas/reference_files/ref_common-1.0.0`
and the common fields.

src/rad/resources/schemas/reference_files/skycells-1.0.0.yaml
```diff
@@ -0,0 +1,199 @@
...
+  skycells:
+    tag: tag:stsci.edu:asdf/core/ndarray-1.*
+    title: Information about all skycells
+    description: |
+      Relevant information about all skycells that have been defined to cover the celestial sphere.
+    datatype:
+      - name: name
+        datatype: [ucs4, 16]
+        byteorder: little
+      - name: ra_center
```
nden, line 156: Is it possible to add a `description` field to these?

src/rad/resources/schemas/reference_files/skycells-1.0.0.yaml
perrygreenfield: Unfortunately `ref_common` has an attribute that is nonsensical for this file, such as `instrument.detector` (there is no detector that is unique to this reference file).

src/rad/resources/schemas/reference_files/skycells-1.0.0.yaml
perrygreenfield: I don't think so, I don't see it as an attribute permitted for the datatype, but I'll ask @braingram  to be sure.

PaulHuwe (approved):
LGTM

src/rad/resources/schemas/reference_files/skycells-1.0.0.yaml
perrygreenfield: The decision has been made to put all the field definition descriptions in the description field for the table as one text block, and that we will ask the appropriate instrument scientist to provide those.

Issue #577, opened 2025-04-10
Schema workflow improvements


braingram:
This issue is to discuss schema/repository workflow improvements.

Viewing schema diffs
------
New schemas are often largely copies of previous schemas. Typical PR tools/workflows are often unhelpful here (since they highlight file diffs and for copies there's nothing to diff).
One option would be to split the schema copy and the changes into different PRs so that the second can be reviewed as a diff. This approach would also work if the changes were separated by commits and the PR was reviewed commit-by-commit (instead of as a total of all changes).

Non-versioned breaking changes
-------
https://github.com/spacetelescope/rad/pull/560 introduced `ref_file-1.1.0` which is used by several other schemas including `wfi_image-1.1.0` (which that PR also introduced). This required new regtest files (since dev romancal uses the dev rad tags/schemas). https://github.com/spacetelescope/rad/pull/574 introduced a `l2_cal_step-1.1.0` which is also used by `wfi_image`. Since `wfi_image-1.1.0` was not yet released it was modified. However this meant that all the L2 files used for regression tests that were valid for #560 now failed validation for #574.

This will be updated as more improvements are discussed.

PaulHuwe:
This isn't a bad idea, but maybe not enough? Desync could still happen similar to what happened with #570 / #571 and #574. e.g.

1. Person A start branch with changes that do not incur a version increment.
2. Person B submits a PR with version increments (straight copies). 
3. Person A submits a PR with their changes & rebases, without realizing new versions exist of their PR.
4. Person B submits a new PR, not realizing that their incremented versions have been desynced due to Person A's PR.

Maybe we could add a check / test where once an advanced (in version number) schema exists, an older version cannot change? This would alert developer and reviewers to a potential desync (would show in step 3 above).

schlafly:
Yes, that's a good point, and I agree also that your suggestion would address it.

WilliamJamieson:
Locally, you can just run your favorite diff program on the between the the new file and the old file. VScode and vimdiff do a pretty good job for this.

This thread may have useful comments: https://stackoverflow.com/questions/30555608/github-compare-how-to-diff-two-different-files-different-file-names-both-in-h

WilliamJamieson:
You can use GitHub Codespaces:

![Image](https://github.com/user-attachments/assets/ceacdbb3-90cb-47be-8da5-30a6973b8e45)

You can then select one of the files for compare:

[...]

PR #574, opened 2025-04-09
RAD-207: Update L1 and L2 Schema

PaulHuwe:
Resolves [RAD-207](https://jira.stsci.edu/browse/RAD-207)

Closes #561 #568 

This PR adjusts several RTB directed L1 & L2 metadata schema changes.

## Tasks
- [x] Update or add relevant `rad` tests.
- [x] Update relevant docstrings and / or `docs/` page.
- [x] Does this PR change any schema files?
  - [x] Schema changes were discussed at RAD Review Board meeting.
- [x] Does this PR change any API used downstream? (If not, label with `no-changelog-entry-needed`.)
  - [x] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).
  - [x] Start a `romancal` regression test (https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) with this branch installed (`"git+https://github.com/<fork>/rad@<branch>"`).
  - [x] Update relevant `roman_datamodels` utilities and tests.

<details><summary>News fragment change types:</summary>

- ``changes/<PR#>.feature.rst``: new feature
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.doc.rst``: documentation change
- ``changes/<PR#>.removal.rst``: deprecation or removal of public API
- ``changes/<PR#>.misc.rst``: infrastructure or miscellaneous change
</details

PR #578, opened 2025-04-10
Base changes to demonstrate a more reviewable schema structure

WilliamJamieson:
This PR reorganizes the rad to follow what was proposed in https://github.com/spacetelescope/rad/issues/577#issuecomment-2795088554. Essentially, this splits the `schemas` directory into two directories `schemas` and `latest`. The `latest` directory is where we should be making changes, while the `schemas` directory is where a schema is moved when it is "frozen". 

Note that the `latest` directory drops the version number from the file name, but maintains that number in the schema's `id` field. This change was necessary to get `git` to show the diff on the schema being updated rather than showing a brand new file. Basically, with the name change and the "copy" being made `git` assumes the original has been renamed to the "copied" file's location and the updates to that schema are treated as a new file.

I have ported the schema changes in PR #574 to this "reorganization". To see the diff it produces see: 

> https://github.com/WilliamJamieson/rad/compare/reorg_to_latest...WilliamJamieson:rad:l2_l2_demo

In my opinion, this diff on GitHub is much easier to follow than the current one. I think I can build a few automation tools to automatically update versions and detect when a schema change requires a new schema. This however is future work.

## Tasks

</details


WilliamJamieson:
There is still work to be done on this PR, but I wanted to "open" it to encourage discussion and to notify maintainers.

WilliamJamieson:
Some additional thoughts:

1. It maybe a good idea to move the "latest" manifest schema into the `latest` directory along side the latest schemas.
2. It may also be a good idea to match the version number of the latest manifest, with that of the `rad` package. This should make it much easier to detect if a new schema needs to be created via the unit tests and some scripting.

jbrookens (approved):
No RDADS updates.

braingram:
I'm in favor of using this strategy of maintaining a "latest". It makes the PR diffs easier to review.

I have one suggest for a further improvement/simplification. If every "latest" file was pointed to by a symlink in the "schemas" (or "manifests") directory with the symlink name matching the current "id" (for example see this [branch](https://github.com/braingram/rad/tree/latest_symlink) with the [latest](https://github.com/braingram/rad/blob/latest_symlink/src/rad/resources/latest/wfi_science_raw.yaml) file and [symlink](https://github.com/braingram/rad/blob/latest_symlink/src/rad/resources/schemas/wfi_science_raw-1.1.0.yaml)) there would be no need to open the schemas to determine the id. When the wheel is built the symlinks are resolved and the installed resources can all use the standard `DirectoryResourceMapping`.

The workflow for updating a schema version would look something like:
- resolve the current symlink to a regular file (to keep the old schema version)
- modify latest, increasing the version number
- create a new symlink

We would benefit from some tests to check:
- does every "latest" have a symlink
- do the symlinks match the schema ids (although any existing "id" matches registered uri check would catch this)

WilliamJamieson:
> I'm in favor of using this strategy of maintaining a "latest". It makes the PR diffs easier to review.
> 
> I have one suggest for a further improvement/simplification. If every "latest" file was pointed to by a symlink in the "schemas" (or "manifests") directory with the symlink name matching the current "id" (for example see this [branch](https://github.com/braingram/rad/tree/latest_symlink?rgh-link-date=2025-04-17T15%3A50%3A37Z) with the [latest](https://github.com/braingram/rad/blob/latest_symlink/src/rad/resources/latest/wfi_science_raw.yaml?rgh-link-date=2025-04-17T15%3A50%3A37Z) file and [symlink](https://github.com/braingram/rad/blob/latest_symlink/src/rad/resources/schemas/wfi_science_raw-1.2.0.yaml?rgh-link-date=2025-04-17T15%3A50%3A37Z)) there would be no need to open the schemas to determine the id. When the wheel is built the symlinks are resolved and the installed resources can all use the standard `DirectoryResourceMapping`.
> 
> The workflow for updating a schema version would look something like:
> 
> * resolve the current symlink to a regular file (to keep the old schema version)
> * modify latest, increasing the version number
> * create a new symlink
> 
> Here's an example: [48173ef](https://github.com/spacetelescope/rad/commit/48173efdd781152094207ba230512c7c96e8853d)
> 
> We would benefit from some tests to check:
> 
> * does every "latest" have a symlink
> * do the symlinks match the schema ids (although any existing "id" matches registered uri check would catch this)

I think this is a reasonable suggestion. I'll make updates to reflect this.

I won't update this to be inline with the main RAD release until we decide to make these changes simply because I have a nice set of branches that demo how the diff looks (linked to in the initial comment).

schlafly:
Let's discuss this on Thursday, but this sounds good to me, thank you!  Do I understand correctly that we could just start doing this now?

WilliamJamieson:
> Let's discuss this on Thursday, but this sounds good to me, thank you! Do I understand correctly that we could just start doing this now?

Yes, all I need to to is update this to the latest version of RAD (I have a local branch that I have kept up to date), I have not simply so that the demo for how changes will look on GitHub still works.

WilliamJamieson:
Note, this should NOT be squash merged. That way the git history is preserved for the original files.

PR #593, opened 2025-05-02
Versioning Assistance App

WilliamJamieson:
This PR adds a terminal application to assist developers with maintaining the versioning of the RAD schemas. This is a follow on to #586.

# Overview of the App

This app is based off the (`textual`)[https://textual.textualize.io] package, which will need to be installed. This dependency is now included as part of the `test` dependencies target. Note that this app is intended only for RAD development and will not be included in any published wheels.

To start the app run the following from the root of your local RAD repo:

```shell
python scripts/rad_version.py
```

This may take a moment to start up since the app is rather inefficient at startup meaning that it will load every single schema and iterate through the RAD git history to compute various details; however, once started it should perform well.

Once started up you will land on a page that looks similar to:

MainPage

The app should be fully usable with a keyboard and/or mouse (so long as your terminal app supports mouse input). Observe that the `<tab>` key will shift the focus (highlighted areas) through all the intractable elements and the `<enter>` key will activate any switches or buttons.
 
Note that the `Quit` button in red, does exactly as advertised; namely, it quits the application returning you to your normal terminal session. Moreover, hitting the `<q>` key whenever a text entry is not focused will also cause the application to quit (even when no quit button is visible). Also note that similarly hitting the `<d>` key will toggle the app's dark mode (default is dark).

Now the app itself is designed to provide three core functionalities:

1. An overview of the `latest` latest resources showing what can and cannot be edited without a version change.
2. Bump the version of any resource in `latest` along with guiding you through the cascade of additional version bumps that will follow.
3. Allow one to create a minimal new schema properly wired into RAD.

## Overview Functionality

As shown in the above image, the app's main page displays an interactive directory tree (scrollable) pointing at the `latest` directory. Each item has one of three icons:

- 📂/📁 indicating a directory (clicking on a directory will toggle the expansion of the contents)
- 📄 indicating a schema that does not need a version bump in order to be edited.
- 🔒 indicating a schema that needs a version bump prior to editing.

## Bumping Schema versions

To bump a schema version select a schema in the directory tree. If the schema needs to be bumped in order to edit it, the blue `Bump` button at the top of the page will be enabled; otherwise, the button will remain disabled.

Clicking `Bump` button will change you to a new screen which should look similar to:

BumpSchema

Notice that there are two buttons:

1. A green `Bump Schema` button (which should be deactivated when you first get to the page).
2. A red `Return` button which returns you to the main page without any changes to RAD.

And there is a scrollable list of new version number text input boxes each labeled with a URI.

In order to enable the `Bump Schema` button one will need to fill in a new and valid version number into each of the input boxes (they can be `<tab>`ed through). Note that by "valid" version number in this context is something that follows:

1. The standard "simple" (3 number) semantic version scheme `a.b.c` where `a`, `b`, and `c` are all integers.
2. The entered version number is strictly greater than the current version. 

When a valid version number for the bump is entered the perimeter of the input box will turn blue, and look something like:

VersionSuccess

While an invalid version number the input box perimeter will turn red, and look like: 

VersionFail

Once the `Bump Schema` button is enabled, 

BumpEnabled

clicking it will bump all the schemas listed and update all the URIs within the resource bodies to match the new bumped URI pattern. You will then be returned to the main page. Notice that on the main page all the schemas that have been bumped are no longer marked as locked:

NewMain

Note that you can bump as main schemas as you want without quitting and reloading the app.

## Creating a New Schema

At any time on the main page you can click the green `Create New Schema` button. This will take you to a new screen that looks something like:

NewSchema

Just as with the Bump Schema screen, the red `Return` button will return you to the main screen with no changes made.

Notice that when you first arrive on this screen 

- Green `Create New Schema` button
- `Tagged:` switch
- `Description:` box

are all disabled. These will be enabled once the correct information is provided.

First one needs to enter in a "valid URI suffix", which is:

1. The suffix indicating the new schema that will be added to the standard RAD URI prefix.
2. The `-` only appears between the schema name and version number.
3. The version number is a valid semantic version number.
4. The URI is not used anywhere in RAD.

Just as the bump schemas, the input box perimeter will remain red until a valid entry is made, at which point it will turn blue. Next ,one fills in a title for the schema. After this the `Tagged:` switch (defaulting to the *off* position) and the green `Create New Schema` button will be enabled. Clicking the `Create New Schema` button will simply create a new un-tagged schema file in `latest` wired correctly into RAD as dictated by the URI.

If you wish to "tag" the schema (generate a registered Tag-URI in the RAD manifest), simply set the switch  to the *on* position. This will enable the description input box (we test that each tag entry in the manifest has a description, but do not test for a description in the schema files themselves, only a title). It will also disable the `Create New Schema` button until the description is filled in.

Clicking the `Create New Schema` button at this point may send you to the Bump Schema page. This occurs when the manifest version needs to be bumped. If you click return on that screen the manifest will not be bumped and you will return the the New Schema screen you were just at, but it will clear the description and turn off the switch. If you instead go through the process to bump the manifest will be bumped and a new schema file will be created and properly wired into RAD (including the new manifest).

Again the note that the main page will be updated to reflect the changes and one can do as many operations as one wants without quitting the app.

## Tasks

</details
Co-authored-by: Brett Graham <brettgraham@gmail.com>


jbrookens (approved):
No RDADS review needed.

WilliamJamieson:
I have done an overhaul of the interface, so the original comment is outdated. The App's functionality is largely the same.

scripts/helper/_app.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def compose(self) -> ComposeResult:
+        """
+        Compose (render in the App) the App's Main screen layout.
+        """
+        yield Header("RAD Development Assistant")
+
+        # Add the TabbedContent widget
+        with TabbedContent(initial="edit"):
+            yield EditTab(self._manager, id="edit")
```
WilliamJamieson, line 73: Currently, there is only one tab here. However, the tabs are here so that future functionality can be added. Such as:

- Integrating the script from #615 as its own tab.
- Integrating a "remove/make static" function as is introduced by #617

schlafly:
Thanks William.  William and Paul have been using this to help deal with version bumps in rad for roughly one build now, so I think we should get this merged.  I'm happy to approve; Paul, any comments before we do?

WilliamJamieson:
> Could you capture the directions you've written up in the PR into the docs? Outside of that, this is good to go. You can branch that off into another PR, if you feel this one has gotten large enough.

Done

PR #611, opened 2025-05-12
Consistency fixes for `maxLength` and `nvarchar`

### LINKED ISSUE #608: `visit_file_group` is `type == integer` but contains a `maxLength` keyword
https://github.com/spacetelescope/rad/blob/43d58fcf52b948a202b1ac1bbb18c70ab7f92fdb/latest/observation.yaml#L146-L153

WilliamJamieson:
Closes #608

This PR addresses two related but distinct issues.

1. `maxLength` should only appear on the `type: string`, as it does not make sense anywhere else.
2. If `archive_catalog.datatype = nvarchar(N)` (where `N` is some positive integer) and the `type` of the node can be a `type: string`, then that `type: string` (option) should have a `maxLength: N` decoration.

Setting `maxLength` consistently, is important so that the schemas can enforce via JSON-schema the length of the stored values so that the archive's `nvarchar` is not exceeded.

~~Note that this touches the `tvac` and `fps` schemas; however, since the versions have been bumped on those schemas there should be no issue. This also fixed a few minor issues with those schemas that were ignored in the 1.0.0 versions.~~

## Tasks

</details

### DIFF
--- a/latest/rcs.yaml
+++ b/latest/rcs.yaml
@@ -1,7 +1,7 @@
 %YAML 1.1
 ---
 $schema: asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.0.0
-id: asdf://stsci.edu/datamodels/roman/schemas/rcs-1.0.0
+id: asdf://stsci.edu/datamodels/roman/schemas/rcs-1.1.0
 
 title: Relative Calibration System Information
 type: object
@@ -28,8 +28,8 @@ properties:
     anyOf:
       - type: string
         enum: ["A", "B", None]
+        maxLength: 5
       - type: "null"
-    maxLength: 5
     archive_catalog:
       datatype: nvarchar(5)
       destination: [WFIExposure.electronics]
@@ -43,8 +43,8 @@ properties:
     anyOf:
       - type: string
         enum: ["1", "2", None]
+        maxLength: 5
       - type: "null"
-    maxLength: 5
     archive_catalog:
       datatype: nvarchar(5)
       destination: [WFIExposure.bank]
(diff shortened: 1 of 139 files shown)


WilliamJamieson:
The `roman_datamodels` issues will be resolved by spacetelescope/roman_datamodels#517.

WilliamJamieson:
This PR is build onto top of #617 now since it is relevant to these changes.

jbrookens (approved):

latest/rcs.yaml
```diff
@@ -28,8 +28,8 @@ properties:
     anyOf:
       - type: string
         enum: ["A", "B", None]
+        maxLength: 5
```
PaulHuwe, line 31: Why move this into the `AnyOf`, given that "null" is also a string?

PaulHuwe (changes requested):

latest/rcs.yaml
WilliamJamieson:
```yaml
type: "null"
```
is NOT a string it is the equivalent of `None`.

WilliamJamieson:
```yaml
      - type: string
        enum: ["A", "B", None]
      - type: "null"
```
Is a statement that says value can be one of the following in Python:

- string: `"A"`
- string: `"B"`
- string: `"None"`
- the `None` python singleton --> i.e.`null` in Java

In practice the asdf file will have one of the values:
- `A` or `"A"`
- `B` or `"B"`
- `None` or `"None"`
- `null`

PaulHuwe (approved):
LGTM

PR #613, opened 2025-05-15
add missing tags to l1_detector_guidewindow

braingram:

Closes #

This PR addresses ...

## Tasks

</details

### DIFF
--- a/latest/datamodels.yaml
+++ b/latest/datamodels.yaml
@@ -1,7 +1,7 @@
 %YAML 1.1
 ---
-id: asdf://stsci.edu/datamodels/roman/manifests/datamodels-1.2.0
-extension_uri: asdf://stsci.edu/datamodels/roman/extensions/datamodels-1.2.0
+id: asdf://stsci.edu/datamodels/roman/manifests/datamodels-1.3.0
+extension_uri: asdf://stsci.edu/datamodels/roman/extensions/datamodels-1.3.0
 title: Datamodels extension
 description: |-
   A set of tags for serializing STScI Roman datamodels.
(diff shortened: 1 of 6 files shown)


tests/test_schemas.py
```diff
@@ -318,6 +318,38 @@ def callback(node, nvarchars=None):
     asdf.treeutil.walk(schema, callback)
 
 
+# don't test tvac or fps schemas as they are static
```
WilliamJamieson, line 321: I don't think this is true anymore as we are now versioning schemas, so I see no reason not to fix issues with tvac/fps schemas so long as we bump the version.

schlafly: That's a good question.  I don't currently ever expect anyone to make TVAC/FPS files, so we could make new versions but we wouldn't expect those to ever see use.

WilliamJamieson: I recall @PaulHuwe mentioning that there were some discussions of updating them. If so then it makes sense to keep maintaining and updating the schemas. Also, it makes sense to keep updating the schemas with fixes and tweaks if we expect the current data to be reprocessed at some point.

PaulHuwe: I recall RTB indicating that there may be future TVAC / FPS releases, and we may make new schemas for them. @tddesjardins or @rcosenti-stsci , care to chip in?

braingram: I added the tvac and fps schemas to this test. There's no need to update the schemas as part of this PR (since none of them fail this test). If we're now updating the fps and tvac schemas perhaps it makes sense to also update the other tests that ignore them (like `test_varchar_length`). Although if they're unused perhaps finding a way to not update them makes sense.

braingram: I opened an issue to discuss the tvac/fps schema plan: https://github.com/spacetelescope/rad/issues/616 Let's move this discussion there.

WilliamJamieson (approved):

tests/test_schemas.py
```diff
@@ -49,6 +49,13 @@
 
 REF_COMMON_XFAILS = ("asdf://stsci.edu/datamodels/roman/schemas/reference_files/skycells-1.0.0",)
 
+ARRAY_TAG_XFAILS = (
+    "asdf://stsci.edu/datamodels/roman/schemas/l1_detector_guidewindow-1.0.0",
+    "asdf://stsci.edu/datamodels/roman/schemas/l1_detector_guidewindow-1.1.0",
```
PaulHuwe, line 54: The existence of these two expected fails implies that we want tests run on old schema versions - is this intended?

PaulHuwe (approved):
LGTM with one question

PR #617, opened 2025-05-22
Remove the FPS/TVAC schemas from latest

### LINKED ISSUE #616: Develop plan for TVAC and FPS schemas
These were previously considered "static". If new versions will never be used it may make sense to not create new versions. If new versions will be used it may be worth updating tests that exclude them to include them.

WilliamJamieson:
Closes #616

This PR removes the FPS and TVAC schemas from the `latest` schemas. In addition to this it adds them to a "new" manifest `static` (name can be changed), which denotes that the schemas tagged in it are no-longer going to be updated and will remain as they are. While not strictly necessary (everything will still function), this should become useful if we want to make changes to RDM which partially break handling of the associated static files, in that case we can build a distinct RDM extension which only handles those schemas where we can maintain a minimal functionality for file reads in RDM.

## Tasks

</details


WilliamJamieson:
Maybe instead of `static` we should use some other term such as:

- stale
- fixed
- constant
- invariant

other suggestions are welcome.

In any case, I'm happy to make adjustments if we decide on a better term.

tests/conftest.py
```diff
@@ -67,111 +190,138 @@ def metaschema_uri():
...
+    Get a URI for a RAD schema from the ASDF resource manager.
 
-@pytest.fixture(scope="session", params=_URIS)
-def uri(request):
-    """
-    Get the URIs of the manifest files from the ASDF resource manager.
+    Note
+    ----
+    Since the schemas are versioned, fixed schema versions cannot be modified so they are
+    not returned by this fixture.
```
WilliamJamieson, line 250: Here I made the `schema_uri` test cases only visit the uris in `latest`. This way we don't pick up failures in `test_schemas` which arise from issues in published schemas.

schlafly:
We chatted about this on Thursday.  I think we thought this was good to go.  I'm happy to sign off; anyone else want to weigh in beforehand?

PaulHuwe (approved):
LGTM

PR #630, opened 2025-07-03
RAD-178: Add effective area to photom

### LINKED ISSUE #459: Update photom schema
In the photom schema, we have under `phot_table` an entry for each optical element that currently contains the following:
- photmjsr
- uncertainty
- pixelareasr

I suggest we add to this the following fields (welcome feedback here):
- wavelength
   - title: Throughput Wavelength Array (micron)
   - description: Array of wavelengths in units of microns at which the throughput has been measured.
   - dtype: array
   - items:
      - dtype: float
- throughput
   - title: Throughput Array (unitless)
   - description: Array of unitless throughput measurements. The throughput is the integrated throughput including all elements of the optical chain from the entrance aperture to the detector pixels including reflectivity, bandpass throughput, and detector quantum efficiency. For the spectral elements GRISM and PRISM, this corresponds to the first-order of the dispersed spectrum.
   - dtype: array
   - items:
      - dtype: float

One thing I have not figured out yet is how to store the additional zeroth-order throughput of the grism which we additionally have. Any suggestions for that are welcome. Maybe an optional field `throughput_zero` ?

tddesjardins:
Resolves [RAD-178](https://jira.stsci.edu/browse/RAD-178)

Closes #459 

This PR addresses adding:
- collecting area in square meters
- wavelength array in microns for effective area curves 
- effective area curves in square meters
for each imaging filter, the grism (first- and zeroth-order), and prism (first-order).

Based on a conversation with @braingram this also reorganizes the patternProperties regex to enforce that imaging filters must have populated the zeropoint, zeropoint uncertainty, nominal pixel area, and effective area information. The GRISM (first-order), GRISM_0 (zeroth-order), and PRISM must have "null" values for the zeropoint, zeropoint uncertainty, and nominal pixel area, but do populate the effective area information. For DARK and NOT_CONFIGURED, all information is "null."

## Tasks

</details

### DIFF
--- a/changes/630.feature.rst
+++ b/changes/630.feature.rst
@@ -1,0 +1,1 @@
+Add effective area to the photom reference file schema.


latest/reference_files/wfi_img_photom.yaml
```diff
@@ -18,37 +18,165 @@ properties:
...
+            unit: "m^2"
+        required:
+          [
+            photmjsr,
+            uncertainty,
+            pixelareasr,
+            collecting_area,
+            wavelength,
+            effective_area,
+          ]
```
WilliamJamieson, line 133: Instead of repeating things multiple times these can be refactored into a definition at the top of the file, and then mixed with with the specific changes for each type.

WilliamJamieson (changes requested):

WilliamJamieson:
@tddesjardins I pushed some updates reflecting my comments please re-review.

WilliamJamieson:
Failing due to versioning issue caused by the new rad release. I will push a fix for that.

schlafly:
Right.  My concern here is more: what (if anything) do we expect to break when this and the corresponding rdm PR is merged?  Will we also need CRDS updates and romancal updates?

PaulHuwe (approved):
LGTM

schlafly:
Trying to restart the discussion here---I'd like to merge this.  Will existing regtests pass for the moment when we do so?  Traditionally when we've updated reference files we've had to tell the CRDS folks and get them to install the new versions and get RTB to upload new files.  I can't decide if versioning saves us---the old files remain valid and the new files will be valid too.  romancal itself uses only the pixelareasr, conversion_megajansky, and uncertainty fields, which look to me to continue to exist at the same places in the files.

If I'm reading this right, then, we can merge this with no breakage, install the updated rdm on the CRDS server with no breakage, and upload the new files to CRDS with no breakage.  In which case we should go ahead with this now.

But pre-versioning the rad update would break romancal until we got everything installed, etc., so I'd appreciate someone else's sense here!

schlafly:
Maybe relatedly---does the failing roman_datamodels test tell me that versioning does not save us in this case?

WilliamJamieson:
> Maybe relatedly---does the failing roman_datamodels test tell me that versioning does not save us in this case?

This mixin needs to be updated in order to fill in the newly added things. https://github.com/spacetelescope/roman_datamodels/blob/233752f35ee50707d1664ef60f1f0152fa97887b/src/roman_datamodels/stnode/_mixins.py#L212-L232

PR #637, opened 2025-07-11
Revert "Remove deprecated exposure types from schema definitions"

braingram:
Reverts spacetelescope/rad#636

This caused several regtest failures for the scheduled run:
https://github.com/spacetelescope/RegressionTests/actions/runs/16208770638

Until we have a fix for this I vote for reverting.

### DIFF
--- a/changes/636.feature.rst
+++ b/changes/636.feature.rst
@@ -1,1 +1,0 @@
-Remove WFI_DARK, WFI_GRISM, and WFI_PRISM from exposure types.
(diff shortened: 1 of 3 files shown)


PaulHuwe:
@schlafly Do you want to revert this, or leave this in since we are updating regtests & files for the Intermediate release anyway?

schlafly:
Going ahead and reverting this since the tool that automatically makes the new regtest files isn't working (github runner sshfs / apt-get issue?), and we'd like to get the intermediate release out ~now.

Issue #639, opened 2025-07-15
Fix fgs_modes_used nvarchar


stscijgbot-rstdms:
_Issue [RAD-214](https://jira.stsci.edu/browse/RAD-214) was created on JIRA by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe):_

The nvarchar value for fgs_modes_used is set to 20, when it should be nvarchar(max)

braingram:
Fixed in #641

PR #641, opened 2025-07-15
Fix fgs_modes_used nvarchar

### LINKED ISSUE #622: fgs_modes_used data type may be incorrect?
The datatype for fgs_modes_used is listed as an array.  The description gives the field as a "list" of modes used during the exposure.

https://github.com/spacetelescope/rad/blob/994b76b5749419f5d114ca5dc6468a9439de9c93/latest/l1_face_guidewindow.yaml#L33-L53

The allowed values are an enum of strings.  I read this as the correct value for this field is something like ['STANDBY', 'WIM_ACQ_VW'], but @tddesjardins , can you confirm that that's the intent here?

@jbrookens , if that is the intent, I think that the rad schema is correct, but I agree that if instead fgs_modes_used is intended to just be 'STANDBY', then we need to update.

PaulHuwe:
Resolves [RAD-214](https://jira.stsci.edu/browse/RAD-214)

Closes #622 #639 

This PR fixes archive catalog string length for fgs_modes_used.

## Tasks

</details

### DIFF
--- a/changes/641.bugfix.rst
+++ b/changes/641.bugfix.rst
@@ -1,2 +1,1 @@
 Fixed archive catalog string length for fgs_modes_used.
-


PaulHuwe:
> We could drop the `maxLength` here:
> 
> https://github.com/spacetelescope/rad/blob/896b277cca301b1e2da77bdea4e7b6b568bab11b/latest/l1_face_guidewindow.yaml#L52
> 
> 
> (since there is already an `enum` the `maxLength` is redundant and only slows down validation).
> No objection from me if that goes into a different PR. At the moment this one (nicely) does not touch any romancal related stuff so I'll approve it as-is.

Maybe we should make a new issue about removing `maxLength` for all enums?

schlafly (approved):

PR #691, opened 2025-09-02
Remove internal tags

### LINKED ISSUE #657: Do we need so many tags?
RAD has a lot of asdf tags, but does it actually need all of them? (this is related to #649)

The purpose of a tag is to confer a special meaning on the information marked by the tag, for example the time tag in an ASDF file's header tells you that the string recorded is actually a time tells you how to figure out from that string what time it is referring to. In makes sense in that context to have a tag there because that value doesn't act like a string even if it is recorded as one. This follows for lots of other objects that are tagged because it sets expectations for how the value is supposed to actually act.

There are some tags in RAD that make sense; namely, those corresponding to the various datamodels we define in RDM. Basically, if a schema has a `datamodel_name` then it should be tagged. 

As for the rest of the tags in RAD, IMOP those don't need to be directly tagged. The vast majority of the tagged objects in the non-datamodel category in RAD are basically fancy dictionaries. Indeed, all the fields defined in these are all accessible through the mapping interface (the corresponding object is a python Mapping -> basically a dictionary). The object itself is a wrapper around a dictionary with some clever stuff glued on to enable things like using `.` to access the fields so they sort of act like class attributes. Moreover, the majority of these "tagged" objects are transparent subclasses of the the same base class, which itself is used at some points anyways. There is a sprinkling of mixins to add small bits of functionality, but for the most part those are for some weird corner cases.

While these objects being tagged does make them independently serializable I haven't seen much use of that functionality and it would likely be for some specialized case where simply serializing the core dictionary would work well enough. The main contextual information for what and how each field acts can all be inferred from the datamodel tag at the top of the tree without having to reference  all the internal tags.

In my personal opinion we can do away with these tags and simply use `$ref` or directly move the schema's content inside relevant object. The `l3_common` schema follows this pattern rather than breaking itself into a large collection of smaller tagged objects, so we are already doing this to some extent.

Note that I think this might help our schemas be more portable for others to use. Tags are not part of the json-schema specification they are an ASDF construct, while `$ref` is core to the json-schema specification. A lot of our uses of `tag` to reference other things defined inside of RAD is really just a complicated way to do the same thing as a `$ref` if you believe my points above. If we remove the use of those tags one can use a standard json-schema library to assemble and work with the majority of the RAD schema structure. Moreover, the actual tag references within RAD schemas would reduce to references to external things at leaf nodes meaning someone could just replace those with a `type: object` and be mostly fine to start with. Currently, they would need to implement something to wedge the tag system in to json-schema just to start off.

My goal here is to start a discussion about this

WilliamJamieson:
Closes #657

This PR removes all uses of internal tags from RAD.

RDM PR: spacetelescope/roman_datamodels#575
RCAL PR: spacetelescope/romancal#1982

## Tasks

</details

### DIFF
--- a/changes/691.feature.rst
+++ b/changes/691.feature.rst
@@ -1,0 +1,1 @@
+Remove the use of internal tags within RAD schemas and replace with standard schema $refs.


WilliamJamieson:
I think this is the way to go, but the metadata blender in Romancal is giving me issues. In particular this function https://github.com/spacetelescope/romancal/blob/83e698c2febbd349f372e39ed7839b6097f09084/romancal/resample/meta_blender.py#L130-L142

Now that all of these are pure dicts by default, they are skipped by this function. However, modifying to handle this change causes `extra` to be included which is not what we wish.

I am actually surprised that this works because `extra` could easily get converted by the node object if one isn't careful, and if one accesses it as an attribute of the datamodel, then https://github.com/spacetelescope/roman_datamodels/blob/26964c51da31835e0251767e387a394a0fffbbde/src/roman_datamodels/stnode/_node.py#L60-L86 will cause it to appear as a `DNode`. 

The main issue is that without recourse to the schemas there is no good way to distinguish a "extra" added to the model and a one of the ones we expect.

WilliamJamieson:
Note that due to wrapping inside of nodes we can end up with LNodes or DNodes in the data dictionary when asdf serializes now, so `roman_datamodels` needs to handle this. This is the only change on top of what we already did for the tagged scalar support.

WilliamJamieson:
This PR is based off #716 because its support in RCAL requires those changes.

WilliamJamieson:
Regression tests look like their failing: https://github.com/spacetelescope/RegressionTests/actions/runs/18165225603/job/51705798130. However, all the failures are simply expected due to the type changes this will induce.

jbrookens (approved):
I don't believe this affects the archive.

braingram (approved):
LGTM. Thanks!

PR #721, opened 2025-10-03
RAD-217 Super Schema and Archive Data parsers

### LINKED ISSUE #659: Create Archive Data Extractor
_Issue [RAD-217](https://jira.stsci.edu/browse/RAD-217) was created on JIRA by [William Jamieson](https://jira.stsci.edu/secure/ViewProfile.jspa?name=wjamieson):_

The Archive would like an automated mechanism to extract the archive_catalog information from RAD so they can process it to create/maintain the database entries.

WilliamJamieson:
Resolves [RAD-217](https://jira.stsci.edu/browse/RAD-217)

Closes #659

This PR is my initial attempt to parse all of the data model schemas into a "super" schema (all information combined with no `allOf` combiners). This then uses that super schema to extract the data as the archive requests.

Currently, this is a work in progress. I will add tests and CI once @jbrookens gets back to me with feedback on the archive outputs.

Note I do not think we should be saving these outputs as they are entirely derived from the files already in existence. I think we should simply have the CI job save them as artifacts saved by the action runner. One would then simply run that action on github on the appropriate branch to get the output for that version of RAD.

## Tasks

</details


src/rad/_parser/_process.py
```diff
@@ -0,0 +1,112 @@
+from __future__ import annotations
```
WilliamJamieson: This may be better in the archive script as it exists only for that right now.

.github/workflows/archive.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Generate archive information
+        run: |
+          python scripts/archive.py --save_dir="${{ github.workspace }}/archive"
+
+      - name: Upload archive information
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: archive
+          path: ${{ github.workspace }}/archive/
```
WilliamJamieson: A workflow output report will probably make this job a little easier to work with. But I think that is outside the scope of the PR.

latest/SSC/ssc_basic.yaml
```diff
@@ -15,6 +15,23 @@ extName: SSC
...
+    archive_catalog:
+      datatype: nvarchar(1024)
+      destination:
+        [
+          CGIExposure.filename,
+          CGIMosaic.filename,
+          CGIAncillary.filename,
+          Foo.bar,
+        ]
+  filename2:
```
PaulHuwe: I think something here may be garbled? "filename2" doesn't exist in the original and is being directed to the same values in the archive as "filename"

PR #740, opened 2025-10-15
Remove default usage and add test confirming no default

braingram:
While looking into other wok I noticed a single `default` usage in the coordinates schema:
https://github.com/spacetelescope/rad/blob/eddda410e2937b05bd87569fd8f305517ce8ebfa/latest/meta/coordinates.yaml#L15

These are no longer automatically filled by asdf which can make it's presence in the schema misleading.

This PR removes the `default` keyword usage and adds a unit test to check that schemas do not use the `default` keyword.

Regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/18531077680

## Tasks

</details

### DIFF
--- a/changes/740.bugfix.rst
+++ b/changes/740.bugfix.rst
@@ -1,0 +1,1 @@
+Remove unused default from coordinates schemas.


schlafly (approved):
Maybe just saying the obvious, in this particular case the change has no effect since the enum has only one possible value.  I've approved the PR.

But just making sure I understand, the conceptual intent here was to give something like create_fake_data a way to find more sensible defaults?  That does seem like something that could be valuable.

braingram:
> Maybe just saying the obvious, in this particular case the change has no effect since the enum has only one possible value. I've approved the PR.
> 
> But just making sure I understand, the conceptual intent here was to give something like create_fake_data a way to find more sensible defaults? That does seem like something that could be valuable.

"default" has a storied past for asdf. More details can be found: https://www.asdf-format.org/projects/asdf-standard/en/1.4.0/schemas.html#default-annotation
but essentially, with the current asdf core schemas version (1.6.0) having a "default" in the schema will do nothing. 1.6.0 has been the default since asdf 4.0.0 (Nov 2024, FWIW we didn't start versioning roman files until Jan 2025). The current asdf specification discourages using "default" (because of the stored past). The linked page has some description of why asdf moved away from using schemas as the location for default values.

PR #787, opened 2025-12-18
RAD-242: add tests for MOC required metadata

braingram:
[RAD-242](https://jira.stsci.edu/browse/RAD-242)

This completes the RAD porition of RAD-242.

This PR adds a few unit tests to confirm select L2 metadata and data that is expected to be of a specific format for the MOC. Any changes to the format, structure or requirement of these data and metadata require advance coordination with the MOC and these tests are meant to provide a warning to any schema changes that might require coordination.

This PR is only new unit tests so no changelog was added and no regtests will be run. This also does not require archive review as it makes no schema changes.

## Tasks

</details

### DIFF
--- a/tests/test_moc_metadata.py
+++ b/tests/test_moc_metadata.py
@@ -30,6 +30,88 @@ import pytest
 
 _L2_SCHEMA_PATH = Path(__file__).parent.parent.absolute() / "latest" / "wfi_image.yaml"
 
+TRUTH = {
+    "data": {
+        "datatype": "float32",
+        "exact_datatype": True,
+        "ndim": 2,
+    },
+    "dq": {
+        "datatype": "uint32",
+        "exact_datatype": True,
+        "ndim": 2,
+    },
+    "meta.filename": {
+        "type": "string",
+    },
+    "meta.observation.execution_plan": {
+        "type": "integer",
+    },
+    "meta.observation.segment": {
+        "type": "integer",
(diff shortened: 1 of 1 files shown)


tests/test_moc_metadata.py
```diff
@@ -0,0 +1,182 @@
...
+    if not values:
+        raise KeyError(f"{key} not found")
+    if len(values) > 1:
+        raise ValueError(f"{key} maps to multiple values: {values}")
+    return values[0]
+
+
+@pytest.mark.parametrize("subschema", ["data"], indirect=True)
+def test_data(subschema):
+    assert subschema["datatype"] == "float32"
```
WilliamJamieson: All of these checks are hard to check if they are correct and all of these tests are approximately the same. We could reduce all these to a single test that uses a single data dictionary (for example an input file provided by the moc).

```python
CHECKS = {
    "data": {
        "datatype": "float32",
        "exact_datatype": True,
        "ndim": 2,
    },
    "dq": {
        "datatype": "uint32",
        "exact_datatype": True,
        "ndim": 2,
    },
}
```

Then all these individual tests can simply be reduced to a single test, and the tests can be verifiable.

WilliamJamieson: See braingram/rad#5 for what I mean.

WilliamJamieson (approved):

PR #793, opened 2026-01-06
Fix issue when file is deleted in archive script

WilliamJamieson:
In PR #789 a latest file AND its symlink were removed, which exposed a bug in the archive comparison script. This is an oversight as for the SOC schemas never result in a file missing form the `src/rad/resources` directory as we are versioning those schemas. However, the SSC schemas are not versioned so when a file/symlink was removed from `src/rad/resources` ASDF fails to fully resolve the old state of RAD.

This PR fixes this by also targeting the changes to `src/rad/resources`.

## Tasks

</details

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,6 @@ _version.py
 
 # uv
 uv.lock
+
+# archive script default write directory
+archive_dump/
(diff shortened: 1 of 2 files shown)


PaulHuwe (approved):
LGTM

PR #870, opened 2026-04-30
Add dust map column to source catalog table.

mairanteodoro:
#### PR #876 needs to be merged first, then this PR needs to be updated to the same version used in that PR.
---

Resolves [RAD-272](https://jira.stsci.edu/browse/RAD-272)

This PR adds a new column to the source catalog table containing information about the estimated dust extinction in B-V, taken by default from the map of Schlegel, Finkbeiner, and Davis (1998).

## Regression tests
- ~https://github.com/spacetelescope/RegressionTests/actions/runs/25376111589/job/74411539791 (05/05/2026; all passing)~.
- https://github.com/spacetelescope/RegressionTests/actions/runs/25581320355 (05/08/2026; all tests passing).

## Tasks

  - [x] Write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types).
  - [x] Start a `romancal` regression test (https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) with this branch installed (`"git+https://github.com/<fork>/rad@<branch>"`).

</details

### DIFF
--- a/changes/870.feature.rst
+++ b/changes/870.feature.rst
@@ -1,0 +1,1 @@
+Added new dust_map column to source catalog table.
--- a/latest/forced_image_source_catalog.yaml
+++ b/latest/forced_image_source_catalog.yaml
@@ -1,7 +1,7 @@
 %YAML 1.1
 ---
 $schema: asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.1.0
-id: asdf://stsci.edu/datamodels/roman/schemas/forced_image_source_catalog-1.4.0
+id: asdf://stsci.edu/datamodels/roman/schemas/forced_image_source_catalog-1.5.0
 
 title: Forced source catalog generated from a Level 2 file by the Source Catalog Step.
 
@@ -17,7 +17,7 @@ properties:
       Photometry and astrometry computed in the Source Catalog Step.
     allOf:
       - tag: tag:astropy.org:astropy/table/table-1.*
-      - $ref: asdf://stsci.edu/datamodels/roman/schemas/tables/forced_catalog_table-1.3.0
+      - $ref: asdf://stsci.edu/datamodels/roman/schemas/tables/forced_catalog_table-1.4.0
 
 required: [meta, source_catalog]
 flowStyle: block
--- a/latest/forced_mosaic_source_catalog.yaml
+++ b/latest/forced_mosaic_source_catalog.yaml
@@ -1,7 +1,7 @@
 %YAML 1.1
 ---
 $schema: asdf://stsci.edu/datamodels/roman/schemas/rad_schema-1.1.0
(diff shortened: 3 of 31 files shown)


braingram:
rdm failures are due to a bug in manifest sorting where "1.10.0" is treated as less than "1.9.0". https://github.com/spacetelescope/roman_datamodels/pull/661 should fix the issue and currently has this rad branch as a dependency to test the issue is fixed.

schlafly:
I think you added the column to the list of possible columns but not to the required columns for any catalog.  I think we should probably require this for at least the image, mosaic, and multiband source catalogs.  

Because it's needed for photo-zs, the requirement is that it appear in the multiband source catalogs; inclusion in the image & mosaic source catalogs makes sense but isn't a strict requirement.

If the way you end up putting this in the code makes the fields also appear for the forced catalogs, we could require it there too; I don't feel strongly about that.

PaulHuwe:
Can you add regression tests when available?

braingram:
Converted to draft from offline conversation with Mairan that the rcal PR is being finished first before this PR is updated to add the column to the list of required columns.

mairanteodoro:
> I think you added the column to the list of possible columns but not to the required columns for any catalog. I think we should probably require this for at least the image, mosaic, and multiband source catalogs.
> 
> Because it's needed for photo-zs, the requirement is that it appear in the multiband source catalogs; inclusion in the image & mosaic source catalogs makes sense but isn't a strict requirement.
> 
> If the way you end up putting this in the code makes the fields also appear for the forced catalogs, we could require it there too; I don't feel strongly about that.

The current implementation makes `dust_ebv` required for the two intended product families (prompt and multiband):

- `latest/tables/prompt_catalog_table.yaml` now includes a required-column constraint for `dust_ebv` (the added `- not: items: not: allOf: ... pattern: ^dust_ebv$` block).

- `latest/tables/multiband_catalog_table.yaml` has the same new `dust_ebv` required-column constraint.

- `latest/tables/forced_catalog_table.yaml` has no corresponding `dust_ebv` block, so forced products are unaffected.

Product impact from table refs:

- Prompt products (`latest/image_source_catalog.yaml`, `latest/mosaic_source_catalog.yaml`) use `prompt_catalog_table-1.4.0` → `dust_ebv` required there.

- Multiband product (`latest/multiband_source_catalog.yaml`) uses `multiband_catalog_table-1.4.0` → `dust_ebv` required there.

- Forced products (`latest/forced_image_source_catalog.yaml`, `latest/forced_mosaic_source_catalog.yaml`) use `forced_catalog_table-1.4.0` → `dust_ebv` **not** required there.

schlafly:
Thanks Mairan, this looks right to me.

schlafly (approved):

PaulHuwe:
I have created #877 to replace this PR.

PR #885, opened 2026-06-02
Allow null pointing fields for CGI dark/calibration data

### LINKED ISSUE #883: Allow CGI pointing fields to be null
From @agreenbaum: 
CGI can have data with no target (e.g., dark or other calibration data that is not associated with a pointing).

Currently we don't call that out in the pointing definitions:
https://github.com/spacetelescope/rad/blob/main/latest/SSC/CGI/keywords/cgi_pointing.yaml

The preference is that such keywords are still sent on the SSC-SOC interface, but with a value of JSON null.  Update these keywords in RAD to use "anyOf", with one of the values being null.

karlhillx:
Closes #883

CGI calibration data can have no target. Made all 6 pointing fields
nullable via anyOf to match the repo convention.

- RA, DEC, RAPM, DECPM: number | null
- TARGET: string | null (maxLength preserved)
- EQUINOX: number | null

### DIFF
--- a/changes/885.misc.rst
+++ b/changes/885.misc.rst
@@ -1,0 +1,1 @@
+Allow CGI pointing fields to accept null values for dark/calibration data.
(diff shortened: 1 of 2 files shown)


cjarnold (approved):
LGTM, thank you!

jbrookens:
No archive changes. Looks good to me.

PaulHuwe:
Please add a change log.

ejoliet (approved):
Looks ok to me!

PaulHuwe (approved):
LGTM

PR #906, opened 2026-08-07
Remove deprecated pip-install input from build workflow

jhunkeler:
* actions/setup-python 7.0.0 removes pip-install

ref #904


PaulHuwe (approved):
LGTM

